PR #17, opened 2023-07-21
Fix install command

bangbangda:
When executing the native:install command, the default nativephp.php configuration file was not published to the config folder.


putera (approved):

mpociot:
Thank you!

PR #26, opened 2023-07-22
Added the option --installer to InstallCommand and DevelopCommand

SergioGMR:
Added the option `--installer=npm` to the command `php artisan native:install` to manage the installation of the npm packages with other installers.
It's added yarn and pnpm.

### DIFF
--- a/src/Traits/Installer.php
+++ b/src/Traits/Installer.php
@@ -2,6 +2,8 @@
 
 namespace Native\Electron\Traits;
 
+use Symfony\Component\Process\Process;
+
 trait Installer
 {
     protected function installNPMDependencies($installer = 'npm')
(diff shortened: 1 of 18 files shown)


mpociot:
I like the idea.
Could you extract this into a trait so that we can use the option also in the `native:serve` and `native:build` commands?

SergioGMR:
Thanks for the clue, yes I will.

SergioGMR:
Plus to the Trait it's added a switchcase to make more readable the situation.

SergioGMR:
Sorry, this is my first time doing a PR to an OS project.
Now you can use the trait without and not need to have the `--installer=` in the $signature, because if it's not present, the command will use the `installWithNPM()` under the hood.

src/Traits/Installer.php
```diff
@@ -0,0 +1,49 @@
...
+                    $this->installWithYarn();
+                    break;
+                case 'pnpm':
+                    $this->info('Installing NPM dependencies using the pnpm package manager...');
+                    $this->installWithPNPM();
+                    break;
+                default:
+                    $this->info('Installing NPM dependencies using the npm package manager...');
+                    $this->installWithNPM();
+            }
```
ConrDev, line 29: maybe you could change this to the following:
```php
switch ($installer = $this->option('installer')) {
    case 'npm':
        $this->info('Installing NPM dependencies using the npm package manager...');
        $this->installWithNPM();
        break;
    default:
        $this->info("Installing NPM dependencies using the {$installer} package manager...");
        $this->installWithCustom($installer);
}

protected function installWithCustom(?string $installer = null)
{
    $installer ??= $this->option('installer');

    $this->executeCommand("{$installer} install", $this->nativePhpPath());
}
```
This way a user could use custom package managers without having to add code.

You could also make an array with installers and the preffered command, for example:
```php
[
    'npm' => 'npm set progress=false && npm install',
    ...
]
```
This way you could add any command when u want to set a specific install command for a certain package manager

SergioGMR:
As you [suggested](https://github.com/NativePHP/electron/pull/26#discussion_r1272880085), I changed the switch case to an array, plus I added a validator that the option exists in the array and if not, I show you what options there are.

SergioGMR:
I can´t resolve the conflict, because I haven't the "Native\Electron\Concerns\LocatesPhpBinary".

SergioGMR:
At the last commit I added the --installer option to DevelopCommand

mpociot:
Thank you for your work on this @SergioGMR

SergioGMR:
My pleasure!

PR #33, opened 2023-07-24
Platform Selection improved

chrisreedio:
Hi @mpociot, 

This is my first pass at changes to help accomodate Windows.

I struggled to get the formatting right with VSCode.

If you have any tips for ensuring that VSCode obeys the repo's standards, I'd love to improve my process here.

We had discussed this quite a bit before but open to make any changes you need.

This will depend on the change in the electron plugin package to fully build on Windows. 

This is also dependent on the windows binary being in the bin dir. Currently it will just fail in a similar way to before but using my fork package it will build. 

Open to input and can change anything you need!

Thanks!


chrisreedio:
@mpociot,

One area I'm struggling with is conditionally importing the phpBinary in `resources/js/src/main/index.js`.

The import is just `php` on Mac/Linux and `php.exe` on Win32.
```js
import phpBinary from '../../resources/php?asset&asarUnpack'
```
```js
import phpBinary from '../../resources/php.exe?asset&asarUnpack'
```

src/Commands/InstallCommand.php
```diff
@@ -37,19 +37,21 @@ protected function nativePhpPath()
...
+        $nativeBinaryPath = $phpBinPackageDir . 'bin/' . (PHP_OS_FAMILY === 'Windows' ? 'win' : 'mac');
+        $this->info("NativePHP binary path: $nativeBinaryPath");
+        $this->info('Fetching latest dependencies…');
+        Process::path(__DIR__ . '/../../resources/js/')
+                ->env([
+                    'NATIVEPHP_PHP_BINARY_PATH' => base_path($nativeBinaryPath),
+                    'NATIVEPHP_CERTIFICATE_FILE_PATH' => base_path($phpBinPackageDir . 'cacert.pem'),
+                ])
+                ->forever()
+				->tty(PHP_OS_FAMILY != 'Windows')
```
LukeTowers, line 50:
```suggestion
                ->tty(PHP_OS_FAMILY != 'Windows')
```

chrisreedio: Thanks for catching this. 
Open to feedback on ways to improve having VSCode adopt the coding style of the repo.

Still somewhat new contributing to open source.

mpociot:
Don't worry about code style. I will install StyleCI to this repository (and the others) as this is nothing that contributors should be worried about 👍

src/Commands/InstallCommand.php
LukeTowers: No worries, it's an easy mistake to make. Best thing you can do in VS Code is enable "Show invisible characters": https://tosbourn.com/how-to-turn-on-invisible-characters-in-vscode/#:~:text=In%20the%20search%20bar%20on,as%20turning%20off%20highlighting%20altogether.

Generally speaking though @mpociot is correct, the best solution is an automated one so that contributors and maintainers never have to think about (or discuss) the style guide.

src/Commands/InstallCommand.php
```diff
@@ -37,19 +37,21 @@ protected function nativePhpPath()
...
+        $this->info("NativePHP binary path: $nativeBinaryPath");
+        $this->info('Fetching latest dependencies…');
+        Process::path(__DIR__ . '/../../resources/js/')
+                ->env([
+                    'NATIVEPHP_PHP_BINARY_PATH' => base_path($nativeBinaryPath),
+                    'NATIVEPHP_CERTIFICATE_FILE_PATH' => base_path($phpBinPackageDir . 'cacert.pem'),
+                ])
+                ->forever()
+                ->tty(PHP_OS_FAMILY != 'Windows')
+                ->run('npm install', function (string $type, string $output) {
```
chrisreedio, line 51: Should this be reverted to being the 'quiet' install? 

The non-quiet version helped solve a few issues when getting this going.

Would it be acceptable to be a bit more verbose during the Alpha?

mpociot: There's the commented out "isVerbose" if condition, which already checks for that. So if you run `php artisan native:install -v` you would get the verbose output.

src/Commands/DevelopCommand.php
```diff
@@ -11,15 +11,19 @@ class DevelopCommand extends Command
 
     public function handle()
     {
+        // PHP_OS_FAMILY on Windows is "Windows"
+        $phpBinPackageDir = config('nativephp.binary_package');
```
mpociot, line 15: Why should this be configurable? 🤔

src/Commands/DevelopCommand.php
```diff
@@ -11,15 +11,19 @@ class DevelopCommand extends Command
...
+        $nativeBinaryPath = $phpBinPackageDir . 'bin/' . (PHP_OS_FAMILY === 'Windows' ? 'win' : 'mac');
```
mpociot, line 16: I think it makes sense to move this logic into a separate trait, as we need it for all commands that make use of the PHP binary. And when we add Linux support, we then only need to update it in one place.

resources/js/package.json
```diff
@@ -27,7 +27,7 @@
     "@electron-toolkit/preload": "^1.0.3",
     "@electron-toolkit/utils": "^1.0.2",
     "@electron/remote": "^2.0.9",
-    "@nativephp/electron-plugin": "^0.1.0"
+    "@nativephp/electron-plugin": "*"
```
mpociot, line 30: This would need to be reverted as this was only required for local testing

mpociot (changes requested):

chrisreedio:
I'll get these changes made today and resubmitted here in the next hour or
two.

src/Commands/DevelopCommand.php
chrisreedio: Disregard as I just re-read your next review. Rolling it into a trait.

chrisreedio:
@mpociot  I believe I have implemented all requested changes.

Really appreciate you looking over this!

chrisreedio:
Now that the binaries are merged. I'm doing some more testing to ensure all of this works properly. 

Will report back shortly once it works 100%.

chrisreedio:
@mpociot The build is now working again. 

u3lQZip

I will get my sample project updated and linked here a bit later this afternoon.

mpociot:
🎉  Awesome, thank you!
I made some minor naming changes - other than that it looks great!

chrisreedio:
I was surprised that it worked too. I believe in the GUI portions if you try to run a file with no extension, it will ask you what program to open it with but in the CLI world, it appears to just execute it. 

Discovered a minor issue with the final `build` command in terms of mac specific flags. 
@fadeaway and I are working that now. Will submit another PR tonight when we get it fixed but the `serve` command does appear fully working in Windows.

mpociot:
@chrisreedio I noticed some things as well, especially when you want to cross-compile.

Right now, the electron-builder.js detects the OS/arch of the _current_ operating system.
But I could also trigger a Windows or Linux build from my Mac!

That's also why there's a `--arm64` argument.

This means that we can not rely on those environment variables and rather need to pass those to the JS scripts.
I already made the modifications locally and successfully built a Windows exe on my Mac, but it's 1 AM now and I need to get some sleep 😅 
I'll review it tomorrow 👍

chrisreedio:
No worries. Good points on the cross compilations. Have a good night!

PR #34, opened 2023-07-24
Fix native:build on MacOS

LukeTowers:
php.js doesn't exist in this context.


PierreThiollent:
It fixes the bug for me too ! 👍🏻

PR #39, opened 2023-07-24
Remove yarn dependency when using serve command

semiherdogan:
Hello,
I've removed yarn dependency from native:serve command.


mpociot:
Thank you!

PR #44, opened 2023-07-27
Added binary lookup for linux based platform

kalizi:
Hi, following from https://github.com/NativePHP/php-bin/pull/6 I changed the binary lookup to include Linux-based platform!


mpociot:
Great, thank you!

PR #47, opened 2023-07-28
Develop command working on Windows again

chrisreedio:
The latest build had broken the develop command on Windows for @fadeaway and myself. 
The develop command should take the current OS / arch into account since it will be running locally on that system. 
The build scripts pass in flags to override the target platform/arch.

* Added single quote  to editor config

Currently on Windows we the native:serve works everytime and the build is 'working' with no errors but we never see a dist directory or output files. 

That was the case before the work in this PR so there must still be something going on.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -7,6 +7,8 @@ indent_size = 4
 indent_style = space
 insert_final_newline = true
 trim_trailing_whitespace = true
+quote_type = single
+max_line_length = 120
 
 [*.md]
 trim_trailing_whitespace = false
--- a/resources/js/electron-builder.js
+++ b/resources/js/electron-builder.js
@@ -12,46 +12,47 @@ const appUrl = process.env.APP_URL;
 const appAuthor = process.env.NATIVEPHP_APP_AUTHOR;
 const phpBinaryPath = process.env.NATIVEPHP_PHP_BINARY_PATH;
 const certificatePath = process.env.NATIVEPHP_CERTIFICATE_FILE_PATH;
-const isArm64 = process.argv.includes('--arm64');
-const isWindows = process.argv.includes('--win');
-const isLinux = process.argv.includes('--linux');
-const isDarwin = process.argv.includes('--mac');
+
 let phpBinaryFilename = 'php';
+
+// Allows us to map the platform name to the directory name for development mode
+const platformDirectory = {
+    win32: 'win',
+    darwin: 'mac',
+    linux: 'linux',
+};
+
+// These are the available platforms we can build for
+const platforms = ['win', 'mac', 'linux'];
+
 // Default to the current platform for develop mode. Build will pass in an arg that overrides this
-let targetOs = process.platform;
+let targetOs = platformDirectory[process.platform];
 
-switch (targetOs) {
-    case 'win32':
-        targetOs = 'win';
-        break;
-    case 'darwin':
-        targetOs = 'mac';
-        break;
-    case 'linux':
-        targetOs = 'linux';
-        break;
-    default:
-        throw new Error('Unsupported platform: ' + targetOs);
-        break;
-}
-
-if (isWindows || targetOs == 'win') {
-    targetOs = 'win';
-    phpBinaryFilename += '.exe';
-} else if (isLinux) {
-    targetOs = 'linux';
-} else if (isDarwin) {
-    targetOs = 'mac';
-}
-
-// Default to the current arch for develop mode. Build will pass in an arg that overrides this
+// Default to the current arch for develop mode.
+// Build will default to x64 but can be set to arm64
 let binaryArch = process.arch;
-if (isArm64) {
-    binaryArch = 'arm64';
-} else if (isWindows || isLinux) {
-    binaryArch = 'x64';
-} else if (isDarwin) {
-    binaryArch = 'x86';
+
+// If we're building, we need to check for target overrides
+if (isBuilding) {
+    // Check for a target platform flag
+    for (const platform of platforms) {
+        if (process.argv.includes('--' + platform)) {
+            targetOs = platform;
+            break;
+        }
+    }
+
+	// Check for forced ARM build, else default to x64
+    if (process.argv.includes('--arm64')) {
+        binaryArch = 'arm64';
+    } else {
+        binaryArch = 'x64';
+    }
+}
+
+// Add .exe to the filename if we're on Windows
+if (targetOs == 'win') {
+    phpBinaryFilename += '.exe';
 }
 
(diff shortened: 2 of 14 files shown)


.editorconfig
```diff
@@ -7,6 +7,8 @@ indent_size = 4
 indent_style = space
 insert_final_newline = true
 trim_trailing_whitespace = true
+quote_type = single
```
chrisreedio, line 10: VSCode kept wanting to put double quotes which was against the current style of this repo so I added this until we have styleci setup

.editorconfig
```diff
@@ -7,6 +7,8 @@ indent_size = 4
...
+max_line_length = 120
```
chrisreedio, line 11: This may also be controversial but again, it kept most lines as they were and until we get styleci setup, it should help.

resources/js/electron-builder.js
```diff
@@ -16,25 +16,44 @@ const isArm64 = process.argv.includes('--arm64');
 const isWindows = process.argv.includes('--win');
 const isLinux = process.argv.includes('--linux');
 const isDarwin = process.argv.includes('--mac');
-let targetOs = 'mac';
 let phpBinaryFilename = 'php';
-if (isWindows) {
+// Default to the current platform for develop mode. Build will pass in an arg that overrides this
```
chrisreedio, line 20: Aligns the target with the directory structure.

resources/js/electron-builder.js
```diff
@@ -16,25 +16,44 @@ const isArm64 = process.argv.includes('--arm64');
...
+let targetOs = process.platform;
```
chrisreedio, line 21: Set the target to the current OS, if building, an override flag will be sent in.

resources/js/electron-builder.js
```diff
@@ -16,25 +16,44 @@ const isArm64 = process.argv.includes('--arm64');
...
+        throw new Error('Unsupported platform: ' + targetOs);
+        break;
+}
+
+if (isWindows || targetOs == 'win') {
     targetOs = 'win';
     phpBinaryFilename += '.exe';
-}
-if (isLinux) {
+} else if (isLinux) {
```
Garanaw, line 41: These else-if blocks seem redundant now (not the first one as it's setting the binary filename)

chrisreedio: The switch normalizes the `targetOs` from `process.platform`, the if statements handle the override flags passed in during build stages.

RoccoHoward: @chrisreedio Just note when it is Linux, but also Win32... at the same time. (Ubuntu in WSL in Windows)

https://github.com/NativePHP/electron/issues/50

chrisreedio: I don't believe you'd get that as the process.platform will always be the actively running platform. So if running it in WSL, it'd be Linux and if running in powershell (et al) it'd be win32 which gets converted to 'win' for the bin directory lookup.

chrisreedio: I run WSL in Windows 10 and have been working with NativePHP from Windows itself. In the long run, I'd prefer to work in WSL and build for Windows. 

Note that only Windows 11 allows graphical WSL (linux apps). If in windows 10, the native:develop command has to be run from Windows (not WSL (for now)).

mpociot: Could you rename the variables in this case?
I think isWindows, isLinux etc is confusing. It's more like "forcedOS" right? Maybe we can simplify this a bit.

chrisreedio:
I agree on all points. Was thinking the same while working on it but didn't
want to drive a bulldozer through that yet but after thinking on it a day
or two, it can definitely be simplified.

I'll get this done later today or first thing tomorrow.

StoneyEagle:
I got to get the Electron app to install the php dependencies.
By prefixing them with npx and node, windows will properly execute the scripts and not open the electron-builder.js file.

vendor\nativephp\electron\resources\js\package.json
```json
{
  "scripts": {
    "dev": "npx electron-vite dev --watch",
    "postinstall": "node electron-builder install-app-deps",
}

**Comment by chrisreedio** [comment]
I got some more done on this today but had some life stuff and wasn't able to get finished. Will do my best to get the rework on this pushed by end of day for me tomorrow. (around 24h from now)

**Comment by chrisreedio** [comment]
I saw this as a merge conflict in the yarn.lock

```
<<<<<<< main
"@esbuild/win32-x64@0.18.17":
  version "0.18.17"
  resolved "https://registry.npmjs.org/@esbuild/win32-x64/-/win32-x64-0.18.17.tgz"
  integrity sha512-fGEb8f2BSA3CW7riJVurug65ACLuQAzKq0SSqkY2b2yHHH0MzDfbLyKIGzHwOI/gkHcxM/leuSW6D5w/LMNitA==
=======
"@esbuild/darwin-arm64@0.18.17":
  version "0.18.17"
  resolved "https://registry.npmjs.org/@esbuild/darwin-arm64/-/darwin-arm64-0.18.17.tgz"
  integrity sha512-M9uJ9VSB1oli2BE/dJs3zVr9kcCBBsE883prage1NWz6pBS++1oNn/7soPNS3+1DGj0FrkSvnED4Bmlu1VAE9g==
>>>>>>> main
```


I went with the Darwin one to be safe but I don't remember swapping this out.

Maybe @mpociot  or someone can advise and double check me on this?

--- 

I still have yet to get the build command on Windows to create a `dist` directory much less place any output files into it.

@fadeaway looked at this at length too and was unsure how to get it working.

I can can revisit this in the coming days. This at least gets the development working well within Windows itself (not WSL).

File: resources/js/electron-builder.js [hunk 6]
```diff
@@ -12,28 +12,48 @@ const appUrl = process.env.APP_URL;
...
+
+	// Check for forced ARM build, else default to x64
+    if (process.argv.includes('--arm64')) {
+        binaryArch = 'arm64';
+    } else {
+        binaryArch = 'x64';
+    }
 }
 
+// Add .exe to the filename if we're on Windows
```
chrisreedio, line 53: We do this check after so that whether we're building for development on Windows OR targeting the build to Windows, it will set the binary extension.

CharlieEtienne:
Hi, I think I get this behaviour on Windows (no dist folder created and no .exe). See [#177](https://github.com/NativePHP/laravel/issues/177)

I tried to replace my files with these commits changes, but I still don't have a dist folder after building.

I also tried different fixes found on other issues, but nothing does the trick for my case...

mpociot:
Alright...I just checked this PR again and the latest changes fixed the develop command on Windows already which is why I'm closing this. 
Thank you for your PR!

PR #52, opened 2023-07-28
Update electron-builder.js

kpanuragh:
- Added platformType variable to store the current platform type
- Updated isWindows, isLinux, and isDarwin variables to use platformType instead of process.argv.includes()
- Updated phpBinaryFilename variable to use 'php' as the default value if isWindows is true


chrisreedio:
Please see my PR as I handle most if not all of this. https://github.com/NativePHP/electron/pull/47

Please note, the default targetOs should not be 'mac' as the develop command should run natively on the same OS as the developer's OS.

The build scripts pass in the target build OS which then overrides the current platform.

kpanuragh:
@chrisreedio  updated default os

mpociot:
Thank you for the PR @kpanuragh 
I'll wait for @ShaneShipston to update his PR before getting back to your PR as he mentioned that some of this is already covered in it.

kpanuragh:
@mpociot  added additional condition checking for platform detection in mac and windows

mpociot:
Thank you!

Issue #55, opened 2023-07-28
[Bug]: Command failed: Syntax error: "(" unexpected


ArtMin96:
### What happened?

I run `php artisan native:serve` command and it just gives this error:

```bash
start electron app...

Error: Command failed: /var/www/nativephp/vendor/nativephp/electron/resources/js/resources/php/php artisan native:config
/var/www/nativephp/vendor/nativephp/electron/resources/js/resources/php/php: 1: Syntax error: "(" unexpected

    at ChildProcess.exithandler (node:child_process:430:12)
    at ChildProcess.emit (node:events:513:28)
    at maybeClose (node:internal/child_process:1091:16)
    at Socket.<anonymous> (node:internal/child_process:449:11)
    at Socket.emit (node:events:513:28)
    at Pipe.<anonymous> (node:net:322:12) {
  code: 2,
  killed: false,
  signal: null,
  cmd: '/var/www/nativephp/vendor/nativephp/electron/resources/js/resources/php/php artisan native:config',
  stdout: '',
  stderr: '/var/www/nativephp/vendor/nativephp/electron/resources/js/resources/php/php: 1: Syntax error: "(" unexpected\n'
}

API server started on port 4003

Starting PHP server... /var/www/nativephp/vendor/nativephp/electron/resources/js/resources/php/php artisan serve /var/www/nativephp

Making sure app folders are available

Skipping Database migration while in development.
You may migrate manually by running: php artisan native:migrate
```

### How to reproduce the bug

Just follow the installation steps.

### Package Version

0.3.2

### PHP Version

8.2.8

### Laravel Version

10.16.1

### Which operating systems does with happen with?

Linux

ArtMin96:
I noticed that the `native:install` command copies the php files from the mac binary.
![image](https://github.com/NativePHP/electron/assets/29732308/485c9b0b-4783-4824-95a8-e9436c9d5e20)

ArtMin96:
https://github.com/NativePHP/electron/pull/57

PR #57, opened 2023-07-29
Linux extra condition for electron builder.

lotharthesavior:
To begin with, thank you so much for your hard work. This is an a fantasticep for the PHP community!

This is not meant to be the final solution, but it solves the problem pointed out at this issue: https://github.com/NativePHP/electron/issues/55

After briefly checking the code, I cou't find any place to trigger the npm post-installation to have the option `--linux`. Due to that, an extra check for the platform was needed for the initial script to copy the binary from Linux folder. 

The same fix might be needed for Windows.

### DIFF
--- a/resources/js/electron-builder.js
+++ b/resources/js/electron-builder.js
@@ -14,7 +14,7 @@ const phpBinaryPath = process.env.NATIVEPHP_PHP_BINARY_PATH;
 const certificatePath = process.env.NATIVEPHP_CERTIFICATE_FILE_PATH;
 const isArm64 = process.argv.includes('--arm64');
 const isWindows = process.argv.includes('--win');
-const isLinux = process.argv.includes('--linux');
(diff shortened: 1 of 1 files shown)


mpociot:
Thank you very much for taking the time to send a PR.
I'd like to wait before emerging this for @chrisreedio and his work on #47 as I think this will already incorporate this change as well as other windows related changes.
I'll come back to this once #47 is ready to be merged

arnolem:
Thanks @lotharthesavior.
It works perfectly for me on WSL/Ubuntu 22.04

marcelo2605:
Thanks @lotharthesavior! Works on Ubuntu 22.04.2

haiderlatitude:
Can anyone explain to me what exactly to do on Ubuntu. I couldn't understand.

mpociot:
As the Windows PR still is being worked on and the changeset is so small, let's get this merged in now 🤷 

Thanks again!

PR #58, opened 2023-07-29
Fix php path in queue command

phuclh:
Fix php path in queue command. It is pointing to the directory, not the php binary

### DIFF
--- a/src/Commands/QueueWorkerCommand.php
+++ b/src/Commands/QueueWorkerCommand.php
@@ -29,7 +29,7 @@ class QueueWorkerCommand extends Command
     {
         $this->info('Starting NativePHP queue worker…');
 
-        $phpBinary = __DIR__.'/../../resources/js/resources/php';
+        $phpBinary = __DIR__.'/../../resources/js/resources/php/php';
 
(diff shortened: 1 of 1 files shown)


mpociot:
Mh...you're probably coming from Windows?
But in any case, there should be a file called `php` in that directory.

phuclh:
@mpociot no, I am using Mac. When I run the native:queue command manually, it throws an error: php is a directory.

Btw, how do you make the queue command auto run like scheduler? I saw in the source code that had an env variable to skip the queue. I tried to set it to false but nothing happened.

Jonnx:
seeing the same error message on a fresh project on mac osx

```
php artisan native:queue
Starting NativePHP queue worker…
sh: /Users/jonasweigert/Development/Personal/com.cachely.app/vendor/nativephp/electron/src/Commands/../../resources/js/resources/php: is a directory
```

mpociot:
Ah you're right, this was introduced with Windows support. Thanks!

Jonnx:
php artisan native:queue now works with the latest changes and processes queue successfully

PR #60, opened 2023-07-29
Resolve Windows opening config file in default application

ShaneShipston:
Windows 10 and 11 will sometimes open the electron-builder.js file in the default application during native:install or native:serve this ensure it's processed correctly.

I confirmed it has no impact on Mac.

It also needs either #47 or #52 to fully resolve Windows builds.


mpociot:
Ahh...that's what caused those issues. Thank you very much!

PR #63, opened 2023-07-30
fix(linux build commands): add missing 'x64' flag

### LINKED ISSUE #62: [Bug]: Incorrect build architecture for Linux (missing --x64 flag)
### What happened?

I cannot currently do the build for Linux x64 when I run `php artisan native:build linux`, the build process automatically targets the arm64 architecture (cf. How to reproduce the bug).

The expected behavior is that the build process should target the x64 architecture for Linux.

### How to reproduce the bug

1. Use a Laravel app
2. Run the command `php artisan native:build linux`
3. Observe that the build is targeted for arm64 instead of x64.

Here a part of the output when Electron build the app:

```bash
=====================
Building for linux | x64
=====================
updater config { provider: 'spaces', name: null, region: null, path: null }
=====================
=====================
Copied app to resources
<app-path>/dist
=====================
  • loaded configuration  file=<app-path>/vendor/nativephp/electron/resources/js/electron-builder.js
  • writing effective config  file=<app-path>/dist/builder-effective-config.yaml
  • skipped dependencies rebuild  reason=npmRebuild is set to false
  • packaging       platform=linux arch=arm64 electron=25.3.2 appOutDir=<app-path>/dist/linux-arm64-unpacked
  • skipping "afterSign" hook as no signing occurred, perhaps you intended "afterPack"?
  • building        target=AppImage arch=arm64 file=<app-path>/dist/<app-name>-0.0.1.AppImage
  • building        target=deb arch=arm64 file=<app-path>/dist/<app-name>_0.0.1_arm64.deb
  • adding autoupdate files for: deb. (Beta feature)  resourceDir=<app-path>/dist/linux-arm64-unpacked/resources
```

As you can see, It looks we have an issue during the build process.

---

I checked the scripts section of the file `resources/js/package.json`(on the `main` branch: https://github.com/NativePHP/electron/blob/f50d279754bc18a0dce8f106e8dd9d7ca2765558/resources/js/package.json), It seems the flag `--x64` is not specified on the linux build and publish commands.

```json
{

    "publish:mac": "cross-env npm run build:mac-arm && cross-env npm run build:mac-x86",
    "publish:mac-x86": "cross-env npm run build && cross-env electron-builder -p always --mac --config --x64 -p always",

    "publish:linux": "cross-env npm run build && cross-env electron-builder -p always --linux --config",
    "build:mac": "cross-env npm run build:mac-arm && cross-env npm run build:mac-x86",
    "build:mac-x86": "cross-env node php.js --x64 && cross-env npm run build && cross-env electron-builder -p never --mac --config --x64",

    "build:linux": "cross-env node php.js --linux && cross-env npm run build && cross-env electron-builder -p never --linux --config"

}
```

In the file `resources/js/package.json`, I suggest adding the `--x64` flag to the linux build and publish commands as following (based on what was done): 
```json
{

    "publish:linux": "cross-env npm run publish:linux-x64",
    "publish:linux-x64": "cross-env npm run build && cross-env electron-builder --linux --config --x64 -p always",

    "build:linux": "cross-env npm run build:linux-x64"
    "build:linux-x64": "cross-env node php.js --linux && cross-env npm run build && cross-env electron-builder -p never --linux --config --x64"

}
```

### Package Version

0.3.2

### PHP Version

8.2.6

### Laravel Version

10.16.1

### Which operating systems does with happen with?

macOS

### Notes

MacOS version 13.4.1

pixrr:
resolve #62

### DIFF
--- a/resources/js/package.json
+++ b/resources/js/package.json
@@ -16,12 +16,14 @@
     "publish:mac": "cross-env npm run build:mac-arm && cross-env npm run build:mac-x86",
     "publish:mac-arm": "cross-env npm run build && cross-env electron-builder -p always --mac --config --arm64 -p always",
     "publish:mac-x86": "cross-env npm run build && cross-env electron-builder -p always --mac --config --x64 -p always",
-    "publish:linux": "cross-env npm run build && cross-env electron-builder -p always --linux --config",
+    "publish:linux": "cross-env npm run publish:linux-x64",
+    "publish:linux-x64": "cross-env npm run build && cross-env electron-builder --linux --config --x64 -p always",
     "build:win": "cross-env node php.js --win && cross-env npm run build && cross-env electron-builder -p never --win --config",
     "build:mac": "cross-env npm run build:mac-arm && cross-env npm run build:mac-x86",
     "build:mac-arm": "cross-env node php.js --arm64 && cross-env npm run build && cross-env electron-builder -p never --mac --config --arm64",
     "build:mac-x86": "cross-env node php.js --x64 && cross-env npm run build && cross-env electron-builder -p never --mac --config --x64",
-    "build:linux": "cross-env node php.js --linux && cross-env npm run build && cross-env electron-builder -p never --linux --config"
+    "build:linux": "cross-env npm run build:linux-x64",
+    "build:linux-x64": "cross-env node php.js --linux && cross-env npm run build && cross-env electron-builder -p never --linux --config --x64"
   },
   "dependencies": {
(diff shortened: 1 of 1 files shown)


mpociot:
Thank you for the PR.
Why is the command called `x86` when it builds `x64`?
Could you rename it?

pixrr:
Yes, of course, I just kept the same naming in other commands 👀

Isn't it a problem to change it for other commands too (mac & win)?

mpociot:
Right, I think that's just an error that we need to fix there. Also the macos PHP binaries still use x86 but should be x64

PR #64, opened 2023-07-31
Correct NATIVEPHP_PHP_BINARY_PATH

lonnylot:
The target OS is set in the `electron-builder.js` file. Having it hard coded here is redundant / leads to a non existent path


mpociot:
Thank you!

PR #68, opened 2023-08-01
Use Node process.platform property to determine the platform

basst85:
Fixes this bug:
https://github.com/NativePHP/laravel/issues/96

### DIFF
--- a/resources/js/electron-builder.js
+++ b/resources/js/electron-builder.js
@@ -12,10 +12,10 @@ const appUrl = process.env.APP_URL;
 const appAuthor = process.env.NATIVEPHP_APP_AUTHOR;
 const phpBinaryPath = process.env.NATIVEPHP_PHP_BINARY_PATH;
 const certificatePath = process.env.NATIVEPHP_CERTIFICATE_FILE_PATH;
-const isArm64 = process.argv.includes('--arm64');
-const isWindows = process.argv.includes('--win');
-const isLinux = process.argv.includes('--linux') || 'linux' === os.platform();
-const isDarwin = process.argv.includes('--mac');
+const isArm64 = process.platform.includes('arm64');
+const isWindows = process.platform.includes('win32');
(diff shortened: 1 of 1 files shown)


mpociot:
Unfortunately we can not fix it like this, as this would loose cross compilation abilities. 

For example, you could create a Linux binary from windows. 
Or a Windows binary on a Mac. 

That's why we have to pass the OS as an argument instead. 

I think the correct way to fix this would be in the "InstallCommand" and then pass the correct OS flag, as determined by PHP.

basst85:
> Unfortunately we can not fix it like this, as this would loose cross compilation abilities.
> 
> For example, you could create a Linux binary from windows. Or a Windows binary on a Mac.
> 
> That's why we have to pass the OS as an argument instead.
> 
> I think the correct way to fix this would be in the "InstallCommand" and then pass the correct OS flag, as determined by PHP.

You're right.

Agree. 
At this moment the OS argument (--win / --linux)  is not passed (correctly) when running npm install from "DevelopCommand", so it thinks you're running mac in stead of windows or linux.

chrisreedio:
@basst85 
Check out my open PR [here](https://github.com/chrisreedio/electron/pull/5).

I believe it should handle this gracefully.

I'm running Windows (and WSL, not that it matters for this topic) and natively in Windows (via powershell in VSCode) the develop command now opens the App without problems. 

Still working on issues with the build command not outputting files but that will most likely be a separate PR.

basst85:
> @basst85 Check out my open PR [here](https://github.com/chrisreedio/electron/pull/5).
> 
> I believe it should handle this gracefully.
> 
> I'm running Windows (and WSL, not that it matters for this topic) and natively in Windows (via powershell in VSCode) the develop command now opens the App without problems.
> 
> Still working on issues with the build command not outputting files but that will most likely be a separate PR.

Thanks! Missed that PR. 
I will close this PR because it's a duplicate fix :-)

PR #71, opened 2023-08-05
phpBinaryPath() reads from config with default value

akunbeben:
This PR will allow developers to use their own PHP binary.

This will also allow developers to pack their needed extensions on their binary.

### DIFF
--- a/src/Concerns/LocatesPhpBinary.php
+++ b/src/Concerns/LocatesPhpBinary.php
@@ -9,7 +9,7 @@ trait LocatesPhpBinary
      */
     protected function binaryPackageDirectory(): string
     {
-        return 'vendor/nativephp/php-bin/';
+        return config('nativephp.binary', 'vendor/nativephp/php-bin/');
     }
 
     /**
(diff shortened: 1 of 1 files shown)


simonhamp:
I would love this in principle.

The problem is that custom binaries potentially exposes end users to nefariously-crafted PHP builds, which is something I really want to avoid.

I need to give this some more thought

akunbeben:
> I would love this in principle.
> 
> The problem is that custom binaries potentially exposes end users to nefariously-crafted PHP builds, which is something I really want to avoid.
> 
> I need to give this some more thought

Okay, that makes sense. 

I would love to hear some thoughts from you. Please let me know if you already have the thoughts about this

simonhamp:
@akunbeben After giving this considerable thought, I've decided that it's a lot of complexity (and risk) for us to allow custom builds of PHP.

From a support perspective, we really need to control the footprint of what's possible and while this feature doesn't _enable_ folks to create custom builds, it does strike the note that as a framework NativePHP _supports_ custom PHP builds, which I think will send the wrong message.

Custom builds introduce **a lot** of variables, which would make it almost impossible for us to provide speedy support. So we need to enforce a stricter approach for now that requires the PHP builds being used to be the ones provided as part of NativePHP.

Think of it like a warranty sticker: it doesn't prevent anyone from bypassing it, but it draws the line in the sand about what we are prepared to support.

PR #72, opened 2023-08-07
App Icon command

simonhamp:
Allows you to copy your app icon into the appropriate location for Electron to pick it up and use it

Make sure your icon is a square PNG and at least 1024px x 1024px

## Usage
```shell
php artisan native:icon
```

This will look for `resources/icon.png` in your application.

Alternatively, you can specify a path:

```shell
php artisan native:icon public/app-icon.png
```

### DIFF
--- a/src/Commands/IconCommand.php
+++ b/src/Commands/IconCommand.php
@@ -10,7 +10,7 @@ use function Laravel\Prompts\note;
 class IconCommand extends Command
 {
     protected $signature = 'native:icon {path=resources/icon.png : The path to your application\'s icon}';
-    
+
     protected $description = 'A command to set your app icon. By default, it will look for `resources/icon.png`.
         Make sure your icon is at least 1024px x 1024px';
 
@@ -18,6 +18,7 @@ class IconCommand extends Command
     {
         if (! file_exists($path = $this->argument('path'))) {
             error("No icon file exists at `{$path}`");
+
             return Command::INVALID;
         }
 
--- a/src/Commands/IconCommand.php
+++ b/src/Commands/IconCommand.php
@@ -1,0 +1,32 @@
+<?php
+
+namespace Native\Electron\Commands;
+
+use Illuminate\Console\Command;
+use function Laravel\Prompts\error;
+use function Laravel\Prompts\intro;
+use function Laravel\Prompts\note;
+
+class IconCommand extends Command
+{
+    protected $signature = 'native:icon {path=resources/icon.png : The path to your application\'s icon}';
+    
+    protected $description = 'A command to set your app icon. By default, it will look for `resources/icon.png`.
+        Make sure your icon is at least 1024px x 1024px';
+
+    public function handle()
+    {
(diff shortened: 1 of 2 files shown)


mdloop:
I am concerned that with this approach that the change is not persistent, say in Git?

mpociot:
I would prefer if the icon would just automatically be copied to the correct location during build time. 
The default location within the codebase can be configured. 
This way there's no manual step involved to copy the icon - this can be done in the electron builder instead.

mdloop:
> 

That would put this in line with the menu bar icons approach as well I believe.

mdloop:
I guess this begs the question which API to put this method in.

None of the existing facades are suitable that I can think of. Maybe "Dock" but that is not available or semantically sound for all OS.

Maybe

```
use Native\Laravel\Facades\App;
```

As long as that does not get confused with the Laravel app.

mpociot:
My idea would be to just put it into the config file and then copy it automatically. 
Changing the app icon should just be a matter of replacing a file in your resources directory - rather a convention than actual configuration. 

I don't think configuring it via an actual Facade/class makes much sense as you can not change the icon at runtime (cross platform) anyway.

simonhamp:
I agree. I saw the command as a stepping-stone towards that - like, this command could be called from whatever build step command, but opens the door to something more complex than just copying a file, whilst keeping the interface straightforward everywhere

Reasoning: In Tauri, there's a command to actually generate a bunch of icon variations from a single command line call, which puts all those files in the right location - having a command that wraps that and works in a similar fashion across both drivers feels neat

mdloop:
> My idea would be to just put it into the config file and then copy it automatically. Changing the app icon should just be a matter of replacing a file in your resources directory - rather a convention than actual configuration.
> 
> I don't think configuring it via an actual Facade/class makes much sense as you can not change the icon at runtime (cross platform) anyway.

Do you think we should do the same for the menu icon then?

```
<?php

return [

...

  'app_icon' => storage_path(env('NATIVEPHP_APP_ICON', 'app/public/appIcon.png')),

  'menu_bar_icon' => storage_path(env('NATIVEPHP_MENU_BAR_ICON', 'app/public/menuBarIcon.png')),

];
```

mpociot:
> Do you think we should do the same for the menu icon then?

No, as not every app has a Menubar - but every app has an icon. So it feels better to configure it along with the menu bar to me, instead of having it in a separate config file.

mdloop:
> > Do you think we should do the same for the menu icon then?
> 
> No, as not every app has a Menubar - but every app has an icon. So it feels better to configure it along with the menu bar to me, instead of having it in a separate config file.

It could be confusing dealing with application assets in such different ways, even if they are different.

Maybe for menu bar, have an API to enable using the icon, and have the icon asset still be configured in the same way as the app icon?

vassilidev:
somes news ?

simonhamp:
I'm closing this PR in favour of #103.

PR #74, opened 2023-08-15
Allow compiling packages in a CI environment

danjohnson95:
Currently, we're unable to compile a NativePHP application in a CI environment (ie GitHub Actions) because dev/tty is not writable.

This change introduces a new `--ci` flag to the following commands:

- native:install
- native:build
- native:publish

This fixes https://github.com/NativePHP/laravel/issues/179 and allows compiling in a CI environment

### DIFF
--- a/src/Commands/BuildCommand.php
+++ b/src/Commands/BuildCommand.php
@@ -12,7 +12,7 @@ class BuildCommand extends Command
 {
     use LocatesPhpBinary;
 
-    protected $signature = 'native:build {os=all : The operating system to build for (all, linux, mac, windows)}';
+    protected $signature = 'native:build {os=all : The operating system to build for (all, linux, mac, windows)} {--ci : Whether the build is running in a CI environment}';
 
     public function handle()
     {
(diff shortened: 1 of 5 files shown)


LukeTowers:
Would `--no-interaction` be a more standard flag to pass?

danjohnson95:
Nice suggestion, thanks. I've updated the commands to use a `--no-interaction` flag instead

GITSKayDomrose:
[GitHub](https://docs.github.com/en/actions/learn-github-actions/variables#default-environment-variables) and [GitLab](https://docs.gitlab.com/ee/ci/variables/predefined_variables.html) (and probably other) set `CI=true` as global environment variable during pipeline.

Maybe check for this as well?

mpociot:
@GITSKayDomrose do you know how other build tools handle this? 
Should this flag just be set automatically if the environment variable is set to true? Would be interesting to know how it works in other systems.

GITSKayDomrose:
> @GITSKayDomrose do you know how other build tools handle this? Should this flag just be set automatically if the environment variable is set to true? Would be interesting to know how it works in other systems.

Jest has an `--ci` [option](https://jestjs.io/docs/cli#--ci), that is also used by this environment variable and changes how it handles snapshots.  
Other than that i don't know any tool by name, but i believe its a common pattern.

I think there should be a dedicated "no-interaction"-mode that could be triggered with a flag or set automatically with `CI=true`, but i think this is out of scope for this MR.  
So for a quick help this is a good first step.

danjohnson95:
What changes are required to get this merged in?

LukeTowers:
@mpociot any chance we could get this merged? It's blocking the larajobs desktop notifier from being able to be compiled for release ❤️

mpociot:
Oh definitely. Sorry for the delay

PR #75, opened 2023-08-17
Upload to GitHub Releases

danjohnson95:
This PR allows users to publish their compiled application to GitHub Releases.

I'll raise a separate PR in the NativePHP/laravel package to update the nativephp.php config stub and another for the docs


mpociot:
This looks good, thank you! I added a commit on the Laravel PR so we don't leak any credentials!
Thank you

PR #77, opened 2023-08-30
windows extra condition for electron builder.

A3Brothers:
Without this condition, when we run "php artisan native:serve" on a Windows OS, it copies the PHP binary from the Mac folder instead of the Windows folder. This occurs because the "isWindows" condition is false, and we never receive a "--win" argument in the "process.argv" array when serving the app. Comparing "win32" from the "os.platform()" method fixes this issue.


mpociot:
Thank you!

PR #81, opened 2023-10-19
add space between cross-env and node

A3Brothers:
Adding a space between "cross-env" and "node" resolved the error that was occurring when running "php artisan native:build mac" without spaces.


simonhamp (approved):
Thanks!

PR #83, opened 2024-02-06
Fix undefined class Process in ExampleTest.php

LunashaGit:
Process was undefined because it was not imported.
```php
    try {
        retry(12, function () {
            // Wait until port 8100 is open
            dump('Waiting for port 8100 to open...');
            $fp = @fsockopen('localhost', 8100, $errno, $errstr, 1);
            if ($fp === false) {
                throw new Exception('Port 8100 is not open yet');
            }
        }, 5000);
    } catch (Exception $e) {
        Process::run('pkill -9 -P '.$process->id()); //Undefined
        throw $e;
    }
    Process::run('pkill -9 -P '.$process->id()); //Undefined
```
But used here : 

```php
    $process = \Illuminate\Support\Facades\Process::path(base_path())
```

### DIFF
--- a/tests/ExampleTest.php
+++ b/tests/ExampleTest.php
@@ -1,6 +1,7 @@
 <?php
 
 use Symfony\Component\Process\PhpExecutableFinder;
+use Illuminate\Support\Facades\Process;
 
 it('can boot up the app', function () {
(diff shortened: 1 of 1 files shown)


simonhamp (approved):
Thanks!

PR #92, opened 2024-05-04
Use Prompts to request Platform when not provided

vikas5914:


src/Commands/BuildCommand.php
```diff
@@ -23,9 +25,18 @@ public function handle(): void
                 echo $output;
             });
 
+        $os = $this->argument('os');
+
+        if ($os == null) {
```
simonhamp, line 30: Could this be simplified to...?

```suggestion
        if (! $os = $this->argument('os')) {
```

src/Commands/BuildCommand.php
```diff
@@ -23,9 +25,18 @@ public function handle(): void
...
+            $os = select(
+                label: 'Please select the operating system to build for',
+                options: ['win', 'linux', 'mac', 'all'],
+            );
```
simonhamp, line 34: Given the discussion in the thread where this started, it would be great if the _current platform_ was the default so the dev doesn't need to type/choose if they're just building to test locally

simonhamp (changes requested):
This feels like a good quality of life improvement, just needs a little more to make it great

src/Commands/BuildCommand.php
vikas5914: Done.. Please check.

simonhamp (approved):
Lovely! Thanks 🙏🏼

PR #98, opened 2024-06-25
Fixed missing import of select() function from Laravel Prompts

rinodrummer:
I was trying to run the `native:publish` command when I encountered the following error:

```
In PublishCommand.php line 34:

  Call to undefined function Native\Electron\Commands\select()
  
```

So I imported the right function from Laravel Prompts to make it work.


simonhamp (approved):
Thanks!

PR #101, opened 2024-07-31
Add environment variable for php bin directory.

kisuka:
Noticed that the php-bin binary path is hardcoded. This isn't great if people need to compile their own versions for needing specific extensions like Zip or Xml. This allows users to set an environment variable to overwrite that.

### DIFF
--- a/src/Concerns/LocatesPhpBinary.php
+++ b/src/Concerns/LocatesPhpBinary.php
@@ -9,7 +9,7 @@ trait LocatesPhpBinary
      */
     protected function binaryPackageDirectory(): string
     {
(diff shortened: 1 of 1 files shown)


simonhamp:
In principle, I think this is fine, but maybe rather than an env key we can just look at a specific path and then fall back to the vendor path?

It would move where this work is done - it would need to shift into the build process rather than in config/env, but I think that will be better overall

simonhamp (approved):

simonhamp:
I've changed my mind on this. I think this is a sensible approach and is easier to implement, more flexible and more explicit.

PR #102, opened 2024-08-26
Updated @nativephp/electron-plugin to v0.5.1

RobertWesner:
Following [release 0.5.1 of electron-plugin](https://github.com/NativePHP/electron-plugin/releases/tag/v0.5.1) I upped the included version here.

Can this go online as 0.7.1 or do we wait for the next "Update the NPM dependencies" release?
I couldn't find information on the current release cycle and would like to use the new features.

### DIFF
--- a/resources/js/package-lock.json
+++ b/resources/js/package-lock.json
@@ -12,7 +12,7 @@
                 "@electron-toolkit/preload": "^3.0.1",
                 "@electron-toolkit/utils": "^3.0.0",
                 "@electron/remote": "^2.0.9",
-                "@nativephp/electron-plugin": "^0.5.0",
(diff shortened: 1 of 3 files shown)


simonhamp:
@RobertWesner is there a better approach here that doesn't require us to manually update the dependency on each release?

I've been thinking to [remove the electron-plugin package altogether](https://github.com/orgs/NativePHP/discussions/348)...

RobertWesner:
+1 on the merging of both repositories. I can't think of a benefit on keeping them separate.
Feel free to close the issue if that is the next reasonable step.

Is there an ETA for the change?

simonhamp:
I think I'll make this my next challenge... it should be quite an easy one

PR #104, opened 2024-09-06
Fixed Tests

RobertWesner:
There were several issue:
- TTY issues due to it not using `--no-interaction`
- Linux required setting up X11 (Xvfb is already installed)

I also cleaned up the only existing Test.

We got to have that little green badge. The workflows passed on my fork.


simonhamp (approved):
Lovely. Thanks 🙏🏼

PR #109, opened 2024-10-05
Integrate Plugin

RobertWesner:
Implemented the [subpath import suggestion](https://github.com/NativePHP/electron/pull/108#issuecomment-2380614265).

Relates to #108

Both `native:serve` and `native:build linux x64` worked.


RobertWesner:
The electron-plugin tests give following error:
```
TypeError: electron_store_1.default is not a constructor
```

As far as I could reproduce this error existed in the electron-plugin repository before it was integrated. For now they will stay broken since the scope of this pull request is the integration via subpath import, not fixing existing tests.

simonhamp (approved):
Looks great! I have to say I prefer this approach over having to do the gymnastics we were trying in #108

What do you think @gwleuverink?

gwleuverink:
Incredible work! This is a way nicer solution 🙌

I like that the plugin build step can be run from the project root too

RobertWesner:
Thank you both for the kind words! This should work on all platforms but it would probably be wise to test on other systems. As I don't have a macOS device at the moment, I'd appreciate someone confirming this working outside of Linux.

simonhamp:
IT WORKS!

PR #113, opened 2024-10-18
Adding support for NSScreenCaptureUsageDescription

obaid:
Adding support for NSScreenCaptureUsageDescription so that my app can capture system audio / screen recording if needed. This needs to be part of build file.

### DIFF
--- a/resources/js/electron-builder.js
+++ b/resources/js/electron-builder.js
@@ -164,6 +164,8 @@ module.exports = {
                 "Application requests access to the user's Documents folder.",
             NSDownloadsFolderUsageDescription:
                 "Application requests access to the user's Downloads folder.",
+            NSScreenCaptureUsageDescription:
+                "Application requests access to the user's screen capture and system audio",
         },
     },
     dmg: {


simonhamp:
But this doesn't need to be a part of everyone's app...

Instead of making this something that's applied to every app, I'd prefer a mechanism here that allows the developer to choose which grants they're going to need

It could be something that goes in the NativePHP config file

obaid:
@simonhamp Happy to help add that ability. If you have any pointers on how to go about it, that would be helpful.

simonhamp:
@obaid I'm not sure exactly, but I'm thinking perhaps a config key in the `config/nativephp.php` such as `mac_entitlements` which contains an array of simplified names, e.g.

```php
'mac_entitlements' => [
    'NSScreenCaptureUsageDescription' => "Application requests access to the user's screen capture and system audio",
]
```

Then there would need to be some logic in `electron-builder.js` to fetch the values from the PHP config (e.g. using an Artisan command) and merge them with some defaults (the ones already in the `electron-builder.js`)

Does this make sense?

simonhamp:
Closing as it's outdated. Feel free to open a new PR with the relevant functionality as discussed if you've managed to make any progress there.

PR #114, opened 2024-10-20
Add support for custom events fired by clicking on notifications

LukeTowers:
Resubmission of https://github.com/NativePHP/electron-plugin/pull/4


simonhamp (approved):
Seems really reasonable to me! Would you be up for implementing the ['client' side](https://github.com/NativePHP/laravel/blob/main/src/Notification.php#L43-L50) too?

LukeTowers:
@simonhamp possibly, what's needed for that? I haven't been working on any nativephp projects in a while.

simonhamp:
@LukeTowers no worries. It just needs a way to pass the custom event name down through the request to the Electron side.

If you're not actively on this right now, I can pick it up.

PR #123, opened 2024-11-08
Improve child process

SRWieZ:
WIP

-  [x] Fix: Process not stopping (Tree kill process)
-  [ ] Fix: Race condition where restarting process keeps spawning new process
-  [ ] Feat: Can stop persistent process

### DIFF
--- a/resources/js/electron-plugin/src/server/api/childProcess.ts
+++ b/resources/js/electron-plugin/src/server/api/childProcess.ts
@@ -97,9 +97,9 @@ function stopProcess(alias) {
         return;
     }
 
-    if (killSync(proc.pid, "SIGTERM", true) === false) {
-        delete state.processes[alias];
-    }
+    killSync(proc.pid, "SIGTERM", true);
+
+    delete state.processes[alias];
 }
 
 function getProcess(alias) {
--- a/resources/js/electron-plugin/src/server/api/childProcess.ts
+++ b/resources/js/electron-plugin/src/server/api/childProcess.ts
@@ -5,6 +5,7 @@ import { notifyLaravel } from "../utils";
 import { join } from 'path';
 
 const router = express.Router();
+const killSync = require('kill-sync');
 
 function startProcess(settings) {
     const {alias, cmd, cwd, env, persistent} = settings;
@@ -96,7 +97,7 @@ function stopProcess(alias) {
         return;
     }
 
-    if (proc.kill()) {
+    if (killSync(proc.pid, "SIGTERM", true) === false) {
         delete state.processes[alias];
     }
 }
--- a/resources/js/electron-plugin/dist/server/api/childProcess.js
+++ b/resources/js/electron-plugin/dist/server/api/childProcess.js
@@ -4,6 +4,7 @@ import state from '../state';
 import { notifyLaravel } from "../utils";
 import { join } from 'path';
 const router = express.Router();
+const killSync = require('kill-sync');
 function startProcess(settings) {
     const { alias, cmd, cwd, env, persistent } = settings;
     if (getProcess(alias) !== undefined) {
@@ -71,7 +72,7 @@ function stopProcess(alias) {
     if (proc === undefined) {
         return;
     }
-    if (proc.kill()) {
+    if (killSync(proc.pid, "SIGTERM", true) === false) {
         delete state.processes[alias];
     }
 }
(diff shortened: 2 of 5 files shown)


SRWieZ:
Explanation updated, ready to review !

resources/js/electron-plugin/src/server/api/childProcess.ts
```diff
@@ -123,10 +137,10 @@ router.post('/stop', (req, res) => {
     res.sendStatus(200);
 });
 
-router.post('/restart', (req, res) => {
+router.post('/restart', async (req, res) => {
     const {alias} = req.body;
 
-    const settings = getSettings(alias);
+    const settings = {...getSettings(alias)};
```
simonhamp, line 143: I don't understand the value of this - wrapping in an object and then unpacking the object that comes back from `getSettings` seems redundant

resources/js/electron-plugin/src/server/api/childProcess.ts
```diff
@@ -96,8 +100,18 @@ function stopProcess(alias) {
...
     }
 
-    if (proc.kill()) {
-        delete state.processes[alias];
+    // Set persistent to false to prevent the process from restarting.
+    state.processes[alias].settings.persistent = false;
+
+    console.log('Process [' + alias + '] stopping with PID [' + proc.pid + '].');
+
+    killSync(proc.pid, 'SIGTERM', true); // Kill tree
```
simonhamp, line 108: Does this work on Windows?

resources/js/electron-plugin/src/server/api/childProcess.ts
```diff
@@ -96,8 +100,18 @@ function stopProcess(alias) {
         return;
     }
 
-    if (proc.kill()) {
-        delete state.processes[alias];
+    // Set persistent to false to prevent the process from restarting.
+    state.processes[alias].settings.persistent = false;
```
simonhamp, line 104: Yep, I like this. Explicitly stopping it makes sense to forcibly stop 👍🏼

resources/js/electron-plugin/src/server/api/childProcess.ts
```diff
@@ -75,9 +76,12 @@ function startProcess(settings) {
             }
         });
 
+        const settings = {...getSettings(alias)};
```
simonhamp, line 79: Same comment as line 143

simonhamp (commented):
Interesting! I hadn't spotted this until now. I can see it definitely orphans child processes. That's super disappointing - my understanding was that utilityProcess would handle this for us 😞 

> I wonder if we should use [child_process.fork](https://nodejs.org/dist/v16.20.2/docs/api/child_process.html#child_processforkmodulepath-args-options) from node.js directly since we don't use [utilityProcess message ports](https://www.electronjs.org/docs/latest/api/utility-process).

I'd like to continue with utilityProcess. Again my understanding (I could be wrong) is that it offers more than just message ports... but I'm open to changing it if it really does just prove to be more painful overall.

> there are some issues on Stack Overflow and GitHub that had the same problem

Are you able to share links to these?

---

I've left a few comments and questions, but nothing should stop this going out imo - I just need to test it on my machine

SRWieZ (commented):
I commented on why I added wrapping and unwrapping.

> I'd like to continue with utilityProcess. Again my understanding (I could be wrong) is that it offers more than just message ports... but I'm open to changing it if it really does just prove to be more painful overall.

> Are you able to share links to these?

Fine with me. I'm not even sure it will fix our problem; it' just that some of the issues I found were related to `child_process.spawn` was used with `{detached:true}` and I was wondering if having the control over this variable would help us ([issue 1](https://stackoverflow.com/questions/44007825/try-to-kill-a-spawn-process-with-shell-true-and-detached-true-nodejs), [issue 2](https://stackoverflow.com/questions/56049102/how-can-i-spawn-a-child-process-that-starts-an-electron-app-and-then-kill-it-whi))

But then I found [this post](https://stackoverflow.com/questions/72897363/child-process-not-killed-node-js-inside-electron-js-app-windows) saying that the problem comes from `{shell: true}` instead.

I can't find the GitHub issues I was referring to. The ones I found in my history were irrelevant. I must have thought they were before digging deeper.

Need a Node.js/Electron expert here.

resources/js/electron-plugin/src/server/api/childProcess.ts
simonhamp: Ah! That surprises me. Ok, in that case this seems fine

simonhamp (approved):

PR #129, opened 2024-11-12
Fix / Forward native ENV variables to child process

gwleuverink:
This PR applies the fix as suggested in https://github.com/NativePHP/laravel/pull/402 on the electron plugin side.

This way we don't need to forward any variables from the process facade. This also is more in line with how internal processes are handled (like the queue & the php server)

Mind that the native env variables are forwarded to every child process. Not exclusively to artisan commands. This was touched on in PR https://github.com/NativePHP/laravel/pull/402. We could do a separate endpoint for artisan commands, though I'm doubting now we know it concerns a small amount of extra variables. There is no harm (I can think of) in passing them along.

### DIFF
--- a/resources/js/electron-plugin/src/server/api/childProcess.ts
+++ b/resources/js/electron-plugin/src/server/api/childProcess.ts
@@ -3,6 +3,8 @@ import { utilityProcess } from 'electron';
 import state from '../state';
 import { notifyLaravel } from "../utils";
 import { join } from 'path';
+import { getDefaultEnvironmentVariables } from "../php";
+
 
 const router = express.Router();
 const killSync = require('kill-sync');
@@ -14,16 +16,22 @@ function startProcess(settings) {
         return state.processes[alias];
     }
 
+    const defaultEnv = getDefaultEnvironmentVariables(
+        state.randomSecret,
+        state.electronApiPort
+    );
+
     const proc = utilityProcess.fork(
         join(__dirname, '../../electron-plugin/dist/server/childProcess.js'),
         cmd,
         {
             cwd,
-            serviceName: alias,
             stdio: 'pipe',
+            serviceName: alias,
             env: {
                 ...process.env,
-                ...env,
+                ...defaultEnv,
+                ...env
             }
         }
     );
--- a/resources/js/electron-plugin/src/server/api/childProcess.ts
+++ b/resources/js/electron-plugin/src/server/api/childProcess.ts
@@ -16,11 +16,6 @@ function startProcess(settings) {
         return state.processes[alias];
     }
 
-    const defaultEnv = getDefaultEnvironmentVariables(
-        state.randomSecret,
-        state.electronApiPort
-    );
-
     const proc = utilityProcess.fork(
         join(__dirname, '../../electron-plugin/dist/server/childProcess.js'),
         cmd,
@@ -31,7 +26,6 @@ function startProcess(settings) {
             env: {
                 ...process.env,
                 ...env,
-                ...defaultEnv,
             }
         }
     );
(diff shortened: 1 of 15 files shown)


resources/js/electron-plugin/src/server/php.ts
```diff
@@ -165,7 +165,7 @@ function getDefaultEnvironmentVariables(secret, apiPort) {
     NATIVEPHP_STORAGE_PATH: storagePath,
     NATIVEPHP_DATABASE_PATH: databaseFile,
     NATIVEPHP_API_URL: `http://localhost:${apiPort}/api/`,
-    NATIVEPHP_RUNNING: true,
+    NATIVEPHP_RUNNING: 'true',
```
simonhamp, line 168: This change feels weird to me... can you explain why a boolean is suddenly a string and how this ends up being ok?

resources/js/electron-plugin/src/server/api/childProcess.ts
```diff
@@ -14,16 +16,22 @@ function startProcess(settings) {
...
         {
             cwd,
-            serviceName: alias,
             stdio: 'pipe',
+            serviceName: alias,
             env: {
                 ...process.env,
-                ...env,
+                ...defaultEnv,
+                ...env
```
simonhamp, line 34: I'm keen to understand the order of assignment here to make sure the correct values are applied... I think it's critical that we don't allow the developer to override some of the `defaultEnv` values here

Maybe `defaultEnv` needs to come last?

```suggestion
                ...env,
                ...defaultEnv,
```

simonhamp (commented):
I like this

As it's done in the `startProcess` method tho (i.e. applies to every child process started this way), I'm still a little concerned that we'll essentially be "leaking" environment variables to whatever process the developer runs.

Whilst I think they're generally well-named environment variables, there's a risk that they may be used incorrectly or even abused... so I'm inclined to suggest that we don't do this by default for everything and instead have either an explicit method for this or an opt-in flag

SRWieZ:
I agree about env variable that should be passed only to php/artisan processes. 

I also think it's important we send the ini configuration to php processes. 

That's why I suggested a php endpoint in the first place.

gwleuverink:
> I agree about env variable that should be passed only to php/artisan processes.
> 
> I also think it's important we send the ini configuration to php processes.
> 
> That's why I suggested a php endpoint in the first place.

Okay! I'll convert this to draft. I was hacking on this before seeing the conclusion of the other PR 😅 Like a real life race condition

To verify I fully comprehend what we're trying to achieve let me recap. 

1. Add php endpoint that starts a process with php_ini config &  `getDefaultEnvironmentVariables`
2. Make the artisan endpoint use the php endpoint internally
3. Bonus: Tidy up [startQueueWorker](https://github.com/NativePHP/electron/blob/0450e7ab7149f8782fa9e0c280ec61d755a79439/resources/js/electron-plugin/src/server/php.ts#L122) to use [getDefaultEnvironmentVariables](https://github.com/NativePHP/electron/blob/0450e7ab7149f8782fa9e0c280ec61d755a79439/resources/js/electron-plugin/src/server/php.ts#L161)

On the Laravel side:
1. Point the ChildProcess calls to the new endpoint
2. Update tests

Do I have the right idea? I'm wondering if we need two endpoints in that case or just the one? `PHP` that only passes ini settings & `artisan` that passes the native env variables & ini settings.

gwleuverink:
Also @SRWieZ if you like to pick up any of this feel free. Didn't mean to swoop in. Only if you want tho of course, no pressure  🤝

SRWieZ:
> Also @SRWieZ if you like to pick up any of this feel free. Didn't mean to swoop in. Only if you want to of course, no pressure  🤝

No worries! 😌 Already gave my input. You can take it from there. I will be less available for 2 days so fill free to finish this.

gwleuverink:
I think this should do it. But I'd like to do some more testing and maybe clean it up a bit before marking it ready for review

simonhamp:
Side note: you don't need to commit build files in PRs as the build is now done [automatically post-merge](https://github.com/NativePHP/electron/blob/main/.github/workflows/build-plugin.yml)

gwleuverink:
Ready! Tested with child processes & the queue (both things I've touched)

SRWieZ (approved):
I read the code here on GitHub, and everything seems fine to me! 
I didn't test this branch on my app, though. 
If every `ini` and `env` setting is merging well, it's okay for me.

simonhamp (approved):
Great work team!

PR #130, opened 2024-11-14
chore - delete native:queue

JustinLawrenceMS:
This PR removes native:queue from the ElectronServiceProvider and also deletes the class PHP file.  I did a grep of both nativephp/laravel and nativephp/electron and there were no other instances

Closes nativephp/laravel#407


simonhamp (approved):
Lovely

PR #135, opened 2024-11-18
test - demo bug

JustinLawrenceMS:
This is a demo of a bug.  It demonstrates that removing electron-store from state.ts fixes the plugin  tests locally.  The code as it is probably breaks others things, so I'm making this a draft and asking for help.  EDIT*  It appears that the 3 plugin tests failed in the workflow but they pass locally


JustinLawrenceMS:
Closing this because a bug report is more appropriate

PR #138, opened 2024-11-19
Dock goodies

simonhamp:
Adds support for:
- Bouncing the Dock icon
- Setting the badge value
- Changing the icon
- Toggling icon visibility

### DIFF
--- a/resources/js/electron-plugin/src/server/api/dock.ts
+++ b/resources/js/electron-plugin/src/server/api/dock.ts
@@ -1,14 +1,61 @@
-import express from 'express'
-import {app, Menu} from 'electron'
-import {mapMenu} from "./helper";
+import express from 'express';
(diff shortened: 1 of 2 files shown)


simonhamp:
> I saw somewhere that app.dock could be null since it's only for MacOS. So maybe use dock?.show()?

Good point @SRWieZ. But the Dock API (in NativePHP) has been like this since inception and I don't recall seeing any reports of issues... but maybe folks aren't doing much cross-platform stuff yet. I will test it out on Windows before merging this.

---

For your issue, I'm not sure I understand the impact on this PR - that sounds like a problem elsewhere... can you raise it as an issue and maybe put a quick screen recording to demo what you're seeing and how to reproduce?

simonhamp:
Tested on Windows and it's a no-op, so I think this is ok

PR #144, opened 2024-12-17
Rename test

JustinLawrenceMS:
This is a very minor PR.  I just noticed that ExampleTest.php now has greater importance than an example, and I renamed it "InitializationTest.php."  If there is another more suitable name, I am happy to revise.


simonhamp:
@JustinLawrenceMS it's not apparent that this fixes something from the tests here (which are still failing)

Is there something I should be doing to understand more why this is necessary?

JustinLawrenceMS:
Hi, the tests are failing because you're testing an outdated Linux version. They are supposed to fail in that situation. I only made linguistic change. Nomenclature only. If you look at the test, it makes the test easier to read. 'ExampleTest.php' was scaffolded in when you installed Pest, but the test more important now and deserves better nomenclature.

PR #145, opened 2024-12-18
latest.yml overwritten when building multiple architectures

gwleuverink:
This PR addresses two issues:

1.  After the latest script update from `publish:mac-arm` to `publish:mac-arm64` the change was not reflected in the `publish:mac` script
2. When publishing for all mac architectures simultaneously it runs the `publish:mac-arm` & `publish:mac-arm64` in sequence. This caused the `latest-mac.yml` file to be overwritten on the second run.

This second point prevents the auto-update feature to detect new releases. I've changed the `publish:mac` script to build for both architectures in a single process. This fixes the issue. 

Check out [this issue ](https://github.com/electron-userland/electron-builder/issues/5592) on the electron-builder repo for more info.

Arguably the same should be done for the Windows & Linux build commands. Will add that if you are okay with this change @simonhamp


simonhamp (approved):
Wonderful!

PR #146, opened 2024-12-20
close print window before freeing mem

labomatik:
Avoid zombies when printing in silent mode


simonhamp (approved):
Seems reasonable. Thanks

PR #147, opened 2024-12-21
chore: cleaning + major versions upgrade 

SRWieZ:
**Summary of changes**
- Removed old `vue` dependencies
- Removed stall files in `dist/` from previous plugin builds (added rimraf before building plugin)
- Fixed some type return in `window.ts`
- Major version upgrade `npm outdated`:
  - `vite` v4 to v5
  - `typescript` v4 to v5 
  - `stylelint` v14 to v16
  - `prettier` v2 to v3
  - `eslint` v8 to v9 
  - `electron-builder` v24 to v25
  - `electron-chromedriver` v31 to v32
  - `electron-context-menu` v3 to v4
  - `get-port` v5 to v7
  - `electron` v31 to v32

**Every commit is checked with (`native:serve`and `native:build`)** 
- ✅ (typescript, vite, jest, eslint, electron-chromedriver, prettier) 1f38d64bf28fbe67a4a5fcfd503661353b3f007e 
- ✅ ESLint upgraded, migrated but not applied 54b83e0c84dfbb7fbd0a1dbc8e2761a497fb9e68
- ✅ ESM migration: 825c342eb34ba2f8cb55c8ceda9b4df47f674b84 245252f5161f4ac4ec7e2e5ae62be701bb0c5d04
- ✅  Update of `get-port`, `electron`, `electron-store` and `electron-context-menu` 84f86e18857df1411babcaafba981199502413f8 bb723113b637c9e04f8c6e3298d7de858fb87c9c df6fcbf954b1d6ea4a71d848f0dd8072ac3dac57

**Upgrades to the last version of `electron` and `vite` blocked by** 
- `menubar`: https://github.com/max-mapper/menubar/pull/486
- `electron-vite`: https://github.com/alex8088/electron-vite/issues/673

**About the ESM upgrade:** 
More and more packages becomes ESM only. Electron supports ESM since v28. Here is some links to read for a better understanding:
- Pure ESM package: https://github.com/sindresorhus/electron-store/releases/tag/v9.0.0
- Pure ESM Package: https://github.com/sindresorhus/get-port/releases/tag/v6.0.0
- Pure ESM Package: https://github.com/sindresorhus/electron-context-menu/releases/tag/v4.0.0
- Gist explaining the steps to migrate and why it became a nightmare to maintain two versions of the same library. https://gist.github.com/sindresorhus/a39789f98801d908bbc7ff3ecc99d99c
- TypeScript and ESM: https://www.typescriptlang.org/docs/handbook/modules/reference.html
[...]

PR #148, opened 2024-12-22
fix: all test suites

SRWieZ:
Back ports of my finding in #147

Funny thing, I started that other PR to just fix the test suite. Then my curiosity led me on an adventure in the JavaScript ecosystem and its quirks. After that, I was so deep into it that I couldn't stop.

**Summary of changes:**
- [JS] Created a base mock for `electron` in jest
- [JS] Updated tests that are mocking `electron` to extend the base one
- [JS] Fixed jest.config.ts regex
- [JS] Fixes somes unused imports ans dependencies
- [PHP] tests against L11 and PHP 8.3
- [PHP] Fixed test by requiring `php-bin^5.0.1`  instead of `php-bin^5.0.0` with a weird zip.
- [PHP] Fixed `phpstan` test suite along side all the errors it did find
- [PHP] Added `composer qa`


SRWieZ:
ALL green! 🎉 ✅

Ready to be reviewed.

src/Updater/UpdaterManager.php
```diff
@@ -52,7 +52,7 @@ public function provider($name = null)
      */
     public function driver($driver = null)
     {
-        return $this->store($driver);
+        return $this->resolve($driver);
```
simonhamp, line 55: Can you explain this change?

simonhamp (commented):
This looks great! I just have one question:

src/Updater/UpdaterManager.php
SRWieZ: Easy. `store` never existed 

I did searched through git history 😄

PR #151, opened 2024-12-28
Add support for window.show()

curtisblackwell:
This PR adds support for showing windows (I'll open the necessary PR in the NativePHP/laravel repo soon). I tried adding tests to match the dialog and clipboard tests, but I couldn't figure out why I was getting an error when I tried asserting a method was called.

Here's the error in case you have suggestions you'd like me to try:

```sh
FAIL  electron-plugin/tests/endpoints/window.test.ts
  ● Test suite failed to run

    electron-plugin/tests/endpoints/window.test.ts:36:21 - error TS2339: Property 'window' does not exist on type 'typeof CrossProcessExports'.

    36     expect(electron.window.show).toHaveBeenCalled();
```

### DIFF
--- a/resources/js/electron-plugin/src/server/api/window.ts
+++ b/resources/js/electron-plugin/src/server/api/window.ts
@@ -108,6 +108,16 @@ router.post('/hide', (req, res) => {
     return res.sendStatus(200);
 });
 
+router.post('/show', (req, res) => {
+    const { id } = req.body;
+
+    if (state.windows[id]) {
+        state.windows[id].show();
+    }
+
+    return res.sendStatus(200);
+});
+
 router.post('/always-on-top', (req, res) => {
     const {id, alwaysOnTop} = req.body;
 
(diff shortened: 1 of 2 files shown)


simonhamp:
I think the error you're seeing is because `window` isn't a child of `electron` - I think you'll need to mock the [`BrowserWindow`](https://www.electronjs.org/docs/latest/api/browser-window) API.

But bear in mind I don't know much about how Jest works 😅

curtisblackwell:
Turns out it was just the `state` that needed to be mocked for these specific tests and they're now passing.

I spent a long time trying to figure out how to mock `electron.BrowserWindow` correctly as well, but no dice. That'll be necessary to more thoroughly test `window.ts`, but I'm not great with JavaScript mocking and don't really have the patience to figure it out on my own.

simonhamp (approved):
Nice. This is fine

PR #153, opened 2025-01-07
enhancement / address build command papercuts

gwleuverink:
This PR adresses a couple of papercuts I've found in the build/publish commands. 

## Fixed

1. **Composer install & cleanup env file were done before the app was copied to the build directory**
    - When the build command finished, all dev dependencies were removed from the base path, breaking some things locally (and in my native:dev script)
    - When the build exits prematurely, the backup env file would not be restored
    
I've addressed this by porting the app copy logic to a php trait. This way the app is copied over before the electron-builder is executed. The copy step inside the electron-builder has been removed in favor for this approach. This way we don't have to modify any userland code.
  
2. **Files that could contain sensitive information were not removed before build**

During the copy step I account for files and directories that should be skipped. I've used an [internal list here](https://github.com/NativePHP/electron/blob/cc758fd203a612440e873be416e1f7b0d6f68a0b/src/Traits/CopiesToBuildDirectory.php#L16), which merges with the list of globs in the nativephp config file.
  
3. **When building all mac architectures at once the `mac-latest.yaml` was overwritten by the last command**

I've tried fixing that before in PR #145. Sadly this caused the last architecture not to compile with the correct php version
      
I've addressed this by removing the php building step as from the package.json scripts, and rather embed them in the build process by using a beforePack hook. 
  
  
## Other points for your consideration

1. **When  `GITHUB_REPO`, `GITHUB_OWNER`, `GITHUB_PRIVATE` are cleaned up, the publish command cannot upload to github.** 

They are all caught by the `GITHUB_*` wildcard. I assume this assume is true for the other update providers too. Maybe the default list of env keys to be cleaned should be more specific, like: `GITHUB_TOKEN`? (that's the only important one that needs to be cleared anyway)

I'm not sure if this still is an issue after these changes. Since we're not directly updating the projects' main env file anymore.

2. **Logs do not rotate. File size can increase dramatically**

Should we default to `LOG_STACK=daily` when building the app?

3. **Should we also clean the `node_modules` directory when building?**

This can become unwieldy quick! A fresh Laravel install already comes with a **30mb** `node_modules` directory.
Not sure if we want to keep `node_modules/.bin` in this case.. An argument could be made to keep that around.

4. **The x64 mac build is packaged with a x86 php binary**

Not sure if this is by design. `php.js` was invoked with the `--x86` flag, while electron builder was invoked with `--mac --x64` in the `build:mac-x86` script.

I kept this behaviour in, but not sure why it is like this


gwleuverink:
I'm sure this needs more polish, but I'd appreciate a review since I'm changing around quite a bit of stuff @simonhamp 

Don't want to make a wrong turn & have to backtrack. Appreciate it! 🙏🏻

gwleuverink:
Tests don't pass for php 8.1 because I've used some syntax that is not 8.1 compatible.

resources/js/electron-builder.js
```diff
@@ -142,6 +66,27 @@ export default {
...
     ],
+    beforePack: async (context) => {
+        let arch = {
+            1: 'x64',
+            3: 'arm64'
+        }[context.arch];
+
+        // If mac is not arm, use 32bits (x86)
+        // This is how this works now according to package.json scripts,
+        // TODO: Check this. Not sure why mac x64 needs a php binary for x86? @simonhamp
```
simonhamp, line 77: It's just a misnomer. I think I picked x64, but the convention elsewhere is x86... so we need to tie the two together. It's not actually a 32-bit arch, it is 64-bit, it's just notated as x86 🤷🏼

src/Traits/CopiesToBuildDirectory.php
```diff
@@ -0,0 +1,103 @@
...
+
+            $this->removeDirectory($path.DIRECTORY_SEPARATOR.$entry);
+        }
+
+        // Clean up
+        $dir->close();
+        rmdir($path);
+    }
+
+    private function keepRequiredDirectories()
```
simonhamp, line 90: I'm not 100% sure why this is needed... the `storage` directory that the application looks for should be the one in `AppData`... would be good to be clear on _why_ this was / still is necessary

src/Commands/BuildCommand.php
```diff
@@ -55,8 +81,33 @@ public function handle(): void
             }
         }
 
-        $this->info((($publish ?? false) ? 'Publishing' : 'Building')." for {$os}");
+        $this->newLine();
+        intro('Updating Electron dependencies...');
+        Process::path(__DIR__.'/../../resources/js/')
+            ->env($this->getEnvironmentVariables())
+            ->forever()
+            ->run('npm update', function (string $type, string $output) {
```
simonhamp, line 89: I think now that we have the NativePHP Electron plugin built in to this package, we can change this to be `npm ci` - it should mean we end up with more reproducible builds and fewer chance of spurious gotchas

src/Traits/CopiesToBuildDirectory.php
```diff
@@ -0,0 +1,103 @@
...
+namespace Native\Electron\Traits;
+
+use RecursiveCallbackFilterIterator;
+use RecursiveDirectoryIterator;
+use RecursiveIteratorIterator;
+
+use function Laravel\Prompts\intro;
+use function Laravel\Prompts\note;
+
+trait CopiesToBuildDirectory
```
simonhamp, line 12: General note here (which does not mean this has to change!):

while PHP's built-in `Recursive*Iterator` classes are great, it may be clearer to have used something like Symfony Finder...

simonhamp:
> When GITHUB_REPO, GITHUB_OWNER, GITHUB_PRIVATE are cleaned up, the publish command cannot upload to github.

As long as the uploader is using the development `.env` and not the one in the build directory, then this shouldn't be a problem now, right?

> Should we default to LOG_STACK=daily when building the app?

Yes. With a pretty short lifespan, so that logs don't hang around too long.

> Should we also clean the node_modules directory when building?

I think in the majority case, whatever the developer was using it for should have been done _before_ the build occurs, so I think there's a good case for just removing it all.

That said, it's not always so clear. I think we need to offload some of this to the developer to consider and provide appropriate config options to make it easy for them to define the behaviour they require.

I also think `vendor/bin` could be another candidate for being removed - highly unlikely that you'll want stray executables knocking around.

> The x64 mac build is packaged with a x86 php binary

See my note in my review.

simonhamp (commented):
This is great. I haven't tested yet, just providing some feedback as requested

resources/js/electron-builder.js
gwleuverink: I'll make that a separate PR once this one is ready 👍

src/Traits/CopiesToBuildDirectory.php
gwleuverink: You are right. That could simplify this part. Will give it a look

src/Traits/CopiesToBuildDirectory.php
gwleuverink: Good point. I ported this part over from the current implementation. Will check if this is still needed

gwleuverink:
Thanks for the review 👍 

Sidenote. How do you feel about using the Laravel prompts spinner to give the build command some extra oomph?

simonhamp:
Sounds fine but we need to check it falls back to something reasonable on Windows

src/Traits/CopiesToBuildDirectory.php
gwleuverink: It seems this is still required. When starting the app for the first time, these folders will not be created unless a file is in there

simonhamp: Ok cools. Can you add a comment in the code here explaining that?

src/Traits/CopiesToBuildDirectory.php
gwleuverink: I have implemented Symfony Finder in commit https://github.com/NativePHP/electron/pull/153/commits/0dba96326cd0b7622cc56f066d483fa2265bf923 but reverted the actual copy part (kept in recursive directory deleting)

Two reasons for this decision.

1. Symfony filter is applied on each item after file collection is done, while using the DirectoryIterator it's applied while crawling the files.

This resulted in the filter callback being called on every file in a otherwise excluded folder. This callback in turn iterates over every exclude pattern & checks the path. This is noticeably slower, even though the real difference is negligible.

Combining the filter with the `except` and `notPath` methods did not help, they are applied after the filters. The `followLinks(false)` method also doesn't work.

Also, when combining the `except`, `notPath` & `filter` methods I'd have to manually filter out paths, files & glob patterns from the exclude list (which partly comes from the config file)

2. During development I use symlinked composer repositories. Due to point number 1 always entering directories even though they are excluded this caused infinite recursion on those local packages.

The boilerplate I had to write to account for this along with the performance hit due to the difference in filtering behaviour I decided the initial DirectoryIterator approach was a bit leaner and more predicable.

Note that the final commit that has the Symfony copying code does not include all the boilerplate and experimenting I mentioned above.

src/Commands/BuildCommand.php
gwleuverink: added in https://github.com/NativePHP/electron/pull/153/commits/7670441801fe240102c9af1523222e844b259edf

src/Traits/CopiesToBuildDirectory.php
gwleuverink: Copied the original comment here: https://github.com/NativePHP/electron/blob/a2633ac4dae53288524d1faea84e199aedb4fff6/src/Traits/CopiesToBuildDirectory.php#L71

gwleuverink:
I've opened PR https://github.com/NativePHP/laravel/pull/467 on the `nativephp/laravel` repo. If accepted, these two should go together.

gwleuverink:
> As long as the uploader is using the development .env and not the one in the build directory, then this shouldn't be a problem now, right?

Yeah it seems that this works now. Probably because in earlier testing I used the upstream CleanEnv trait without moving the code to the build directory first.

> I think in the majority case, whatever the developer was using it for should have been done before the build occurs, so I think there's a good case for just removing it all.

> I also think vendor/bin could be another candidate for being removed - highly unlikely that you'll want stray executables knocking around.

Added this in PR https://github.com/NativePHP/laravel/pull/467

gwleuverink:
> Sounds fine but we need to check it falls back to something reasonable on Windows

According to the docs prompts should use fallbacks for windows without WSL. https://laravel.com/docs/11.x/prompts#fallbacks

Unfortunately I don't have a machine to test this on handy. Might be able to check with vmware. But if this is important I'll hold off on this for now. Focus on making this rock solid first

gwleuverink:
>> Should we default to LOG_STACK=daily when building the app?

> Yes. With a pretty short lifespan, so that logs don't hang around too long.

I can add this to the service provider on in the Laravel repo. This seems like a good place. There already are some config overrides in there.

One thing to note is that anything that's logged before the serviceprovider loads still ends up in the default laravel.log file. To address this we could insert the log config in the build env file instead. 

It's a bit weird maybe that only this thing is handled differently, but this way we're sure the right log channel is always used immediately when the app boots.

Otherwise if this is not an objection, the service provider should be fine too. Which one would you pick? @simonhamp

src/Traits/CopiesToBuildDirectory.php
simonhamp: Understood! Thanks for doing that, sorry it didn't work out. This is fine

src/Traits/CleansEnvFile.php
```diff
@@ -7,22 +7,35 @@
 
 trait CleansEnvFile
```
gwleuverink, line 8: NOTE: Added the overriding of the LOG env variables here. Committed this with wrong message

gwleuverink:
Almost done. Couple things to note:

1. The Prompts spinner did not look as nice as I'd hoped due to some build steps forwarding process output. (npm ci & composer install)

2. There used to be a php minification step. I dropped that from the electron repo, since it was not used currently. Should I bring that back?

3. Do you think we should cache views during build (and maybe routes & config too?)

SRWieZ:
**First I tested as a user :**

-  ✓ It’s good that `->ask()` is the first thing in the build command.I would push setting the app name even later.
-  ✓ `php artisan native:build mac` doesn’t do "all" by default like before, but at least it's consistent with `php artisan native:build` asking for the OS.  
-  ✓ `php artisan native:build mac all` seems fixed.  
-  ✓ Usages of `intro` make the CLI output much more readable.


**Then as a contributor:**
- ✓ Love the separations of concerns into Trait + the tests. Awesome 🤩 
- ✓ Love that some logic from electron-builder is ported to PHP ^^
- We should PR the change into `native-php` right now for the x86/x64 confusion and take a fresh start for the next release.

SRWieZ:
**My only rant:**

That has objectively nothing to do with your PR, but your cleaning may facilitate the change.

Personally, I’m still bothered that the build directory is in `vendor/`. 

It has caused me a lot of problems when contributing:
-  I can’t easily clean when the Electron download has failed.
-  Composer dev requirements are deleted after every build.
-  When working on a PR with a symlinked library, package.json is modified.
-  It also slows my IDE by indexing a bunch of files that are not in an ignore directory but a resource directory and detects "duplicated classes."
-  `vendor` and `node_modules` are considered read-only directories by tools and IDEs.
-  It will be on par with other cross-platform tools (Cordova uses `platforms/{ios/android}`, Capacitor and React Native use `ios/` and `android/`, Flutter uses `build/{ios/android/web}`).

It would be better if the building process were somewhere in the top directory, let's say `build/` or `temp/`.
-  `resource/js` would be copied first into `build/`.
-  Change `buildPath` to `build/resources/js/resources/app/`.
-  Change `__DIR__.'/../../resources/js/'` references accordingly.
-  Reorder some build steps.

If you find me annoying, you can tell me it's fine 😄

SRWieZ (approved):

gwleuverink:
@SRWieZ thanks for your feedback!

I definetly noticed some ide slowdowns and crashes. Let's pick that up in the next PR.

I'll go through your other points (php-bin & moving app icon) next.

---

There seems to be one test that won't run in windows. But otherwise this PR is good to go

.github/workflows/run-tests.yml
```diff
@@ -20,12 +21,12 @@ jobs:
           - laravel: 10.*
             testbench: ^8.18
             carbon: ^2.67
-            laravel-package-tools: ^1.16
+            laravel-package-tools: ^1.16.4
```
gwleuverink, line 24: The version in the Laravel repo wasn't compatible with the matrix we're running. Needed a bump to prevent composer errors in CI

gwleuverink:
Okay. We're here. 
All of this PR's tests are green. 

**Couple things to check - both unrelated to this PR.** 
- Plugin test & plugin build fail - I believe due to a upstream change
- BootingTest doesn't run in ubuntu

Otherwise I think this is good to go.

SRWieZ:
@simonhamp since you're the original reviewer, I would appreciate your final approval when you have time.

simonhamp (approved):
Looking great!

PR #154, opened 2025-01-13
feat: replace jest with vitest

SRWieZ:
**Replacing Jest with Vitest 🎉**

Works out of the box with ESM and has fewer dependencies. And, error messages are much more clear.

  - Updated tests to use `vitest` instead of. `jest`
  - Added a new mock implementation for the `electron` module using `vitest` and a new test file to verify the mocking of the `electron` module using `vitest`. (`resources/js/electron-plugin/tests/mocking.test.ts`)
  - Removed  a bunch of `jest` configuration files and dependencies since `vitest` supports ESM out of the box.
  - Bumped dependencies of `vite`, `electron-vite` and some `eslint` and `stylelint` plugins 

**But ⚠️ :**

- Removed old Jest mocks and tests for clipboard, dialog, and window endpoints (`endpoints/clipboard.test.ts`, `endpoints/dialog.test.ts`, `endpoints/window.test.ts`).
-  Mocking Electron is unreliable in both Jest and Vitest. I managed to get some tests working, but others broke. I believe we need to shift to end-to-end tests using Selenium/Playwright and the kitchen-sink app. The remaining tests still verify the boot process of the API server.


SRWieZ:
This is the best I can do right now 😅

If you have any suggestion, feel free to edit this PR. 

Updated my first comment.

@gwleuverink @PeteBishwhip @simonhamp

PeteBishwhip (approved):

simonhamp:
Should we get these conflicts resolved and merge this in too?

SRWieZ:
Yes, we should.

I now strongly believe this is not the right way to test electron features/api. They are still a bit useful in this state, but we should rely more on e2e in the future.

SRWieZ:
I can merge conflict later this afternoon.

SRWieZ:
Ready!

simonhamp:
I wonder why the plugin build is failing now here... maybe it's missing the GitHub token in that action?

SRWieZ:
Yeah, I saw, but it's unrelated to my code, the build works, It's committing on my branch that doesn't.

I think it's because this branch is still in my old fork of the project (srwiez/fork-nativephp-electron)

So, I suppose the workflow won't work on our contributors' branches.

Anyway, it seems like this workflow still needs some love.

simonhamp:
>I suppose the workflow won't work on our contributors' branches.

Yep that's it

simonhamp (approved):

PR #156, opened 2025-01-16
fix: menubar tray state

SRWieZ:
Creating a context menu only tray is now fully independent of menubar plugin.


simonhamp (approved):
Very nice

PR #158, opened 2025-01-17
Add Windows support and deep linking configuration

JulianaChiabai:


simonhamp:
Thanks for this @JulianaChiabai! Would you mind just providing a brief description and tag any issues this may fix or discussion threads it relates to?

simonhamp (approved):

gwleuverink:
@simonhamp I'm working on fixing tests/builds on PR https://github.com/NativePHP/electron/pull/153. Just pulled in main and this PR seems to break the `npm plugin:build` for me.

SRWieZ:
I can confirm this. This PR uses this.handleDeepLink(), which doesn't appear to be declared.

It should probably replicate what the macOS code does: 
```js
app.on("open-url", (event, url) => {
  notifyLaravel("events", {
    event: "\\Native\\Laravel\\Events\\App\\OpenedFromURL",
    payload: [url],
  });
});
```

And send `comandline` instead of `url`

I propose to modify `OpenedFromURL` to support both and keep one event for the 3 os
```php

class OpenedFromURL implements ShouldBroadcastNow
{
    use Dispatchable, InteractsWithSockets, SerializesModels;

    public function __construct(public ?string $url = null, public ?string $commandLine= null) {}
```

PR #162, opened 2025-02-05
Update logic to ensure TTY mode is appropriately disabled

PeteBishwhip:
Fixes https://github.com/NativePHP/laravel/issues/468

- Utilises the Symfony Process class which has a tried and tested `isTtySupported()` method. This will return false if windows due to a DIRECTORY_SEPARATOR check.
- Removed check for Windows specifically as this is checked as above.
- Ensures even if TTY is supported, it is still set to off when -n|--non-interactive is specified.

### Supporting Documentation:
https://symfony.com/doc/current/components/process.html#checking-for-tty-support

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -34,7 +34,7 @@
         "php": "^8.1",
         "illuminate/contracts": "^10.0|^11.0",
         "laravel/prompts": "^0.1.1|^0.2|^0.3",
-        "nativephp/laravel": "^1.0",
+        "nativephp/laravel": "^1.0@beta",
         "nativephp/php-bin": "^0.5.1",
         "spatie/laravel-package-tools": "^1.14.0"
     },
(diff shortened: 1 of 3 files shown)


composer.json
```diff
@@ -34,9 +34,9 @@
         "php": "^8.1",
         "illuminate/contracts": "^10.0|^11.0",
         "laravel/prompts": "^0.1.1|^0.2|^0.3",
-        "nativephp/laravel": "^1.0",
+        "nativephp/laravel": "^1.0@beta",
```
SRWieZ, line 37: Do you remember why you needed that ?

PeteBishwhip: As `1.0` isn't tagged in a way that Composer currently sees as `stable` in the laravel repo. Until that time, it needs to be specified to use `@beta` or higher specifically to allow it to retrieve the 1.0 versions.

SRWieZ: Yeah, I think I remember encountering the same issue. 

Let's merge it, and we will change it before the next stable release.

SRWieZ (approved):

gwleuverink (approved):

SRWieZ:
If @gwleuverink approves it, we can merge it. Next, we need to merge Willem's work before making any other changes to the build command.

PR #166, opened 2025-02-17
Support Laravel 12.x & PHP 8.4

SRWieZ:
Replaces #165 

* Support Laravel 12 & PHP 8.4
* Refactored to group feature and unit tests separately, so we can work on E2E tests separately
  * Added MacOS to unit tests
  * Isolated BootingTest into his own workflow

More details in the comment below. 

Depends on:
- https://github.com/NativePHP/laravel/pull/494

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -32,7 +32,7 @@
     ],
     "require": {
         "php": "^8.1",
-        "illuminate/contracts": "^10.0|^11.0",
+        "illuminate/contracts": "^10.0|^11.0|^12.0",
         "laravel/prompts": "^0.1.1|^0.2|^0.3",
         "nativephp/laravel": "*",
         "nativephp/php-bin": "^0.5.1",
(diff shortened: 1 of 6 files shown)


composer.json
```diff
@@ -32,24 +32,24 @@
     ],
     "require": {
         "php": "^8.1",
-        "illuminate/contracts": "^10.0|^11.0",
+        "illuminate/contracts": "^10.0|^11.0|^12.0",
         "laravel/prompts": "^0.1.1|^0.2|^0.3",
-        "nativephp/laravel": "*",
+        "nativephp/laravel": "dev-main",
```
simonhamp, line 37: I know this has probably been done because `nativephp/laravel` with L12 support hasn't been released yet (and that while we have the `-beta` suffix, it's a bit more painful)... so I'm going to leave this here, but just wanted to say that we need to come back and switch this to `^1.0` when the time comes

SRWieZ: Yes absolutely.

PeteBishwhip (approved):

PR #167, opened 2025-02-18
Pre & Post Build Command Support

PeteBishwhip:
Config key defaults provided in https://github.com/NativePHP/laravel/pull/496

This pull request introduces a new trait for handling pre- and post-processing commands in the build process, along with updates to the `BuildCommand` class and associated tests. The most important changes include adding the `HasPreAndPostProcessing` trait, updating the `BuildCommand` to use this trait, and adding tests to ensure the trait works correctly.

### New Trait for Pre- and Post-Processing:

* [`src/Traits/HasPreAndPostProcessing.php`](diffhunk://#diff-379efec112599f6e647c8cd6a05232325a4090cf76207cad607a753eb6298c63R1-R92): Introduced the `HasPreAndPostProcessing` trait to handle pre- and post-processing commands during the build process. This includes methods for executing commands and retrieving configuration.

### Updates to BuildCommand:

* [`src/Commands/BuildCommand.php`](diffhunk://#diff-aa5dd2724f30cebdd7795cf02ae6a572645fd33f0048b4cb27f37fdc985a46b0R10-R16): Updated the `BuildCommand` class to use the new `HasPreAndPostProcessing` trait and added calls to `preProcess` and `postProcess` methods in the `handle` method. [[1]](diffhunk://#diff-aa5dd2724f30cebdd7795cf02ae6a572645fd33f0048b4cb27f37fdc985a46b0R10-R16) [[2]](diffhunk://#diff-aa5dd2724f30cebdd7795cf02ae6a572645fd33f0048b4cb27f37fdc985a46b0R32-R33) [[3]](diffhunk://#diff-aa5dd2724f30cebdd7795cf02ae6a572645fd33f0048b4cb27f37fdc985a46b0R71-R72)

### Tests for the New Trait:

* [`tests/Mocks/HasPreAndPostProcessingMock.php`](diffhunk://#diff-c53680a211f5de4af3d952cf4dfd395d93924e89ce516eea5f4782787733193fR1-R11): Added a mock class to facilitate testing of the `HasPreAndPostProcessing` trait.
* [`tests/Unit/Traits/HasPreAndPostProcessingTest.php`](diffhunk://#diff-6f2e8baf5d362b921881c05c08d350c452293902fa6f803d509eade3e624539eR1-R49): Added unit tests to verify that the `preProcess` and `postProcess` methods execute commands as expected based on configuration.

### Development Dependency:

* [`composer.json`](diffhunk://#diff-d2ab9925cad7eac58e0ff4cc0d251a937ecf49e4b6bf57f8b95aab76648a9d34R43): Added `mockery/mockery` as a development dependency for testing purposes.

Closes [#493](https://github.com/NativePHP/laravel/issues/493)


src/Traits/HasPreAndPostProcessing.php
```diff
@@ -0,0 +1,94 @@
...
+        $config = config($this->formatConfigKey($configKey));
+
+        if (is_array($config)) {
+            // Filter out empty values
+            return collect($config)
+                ->filter(fn ($value) => ! empty($value));
+        }
+
+        return $config;
+    }
```
gwleuverink, line 72: This is personal preference but I feel `getPrePostBuildConfig` & `formatConfigKey` are pretty simple in what they achieve, yet this does not communicate intent well to me personally.

Don't both methods combined come down to these two lines?

``` php
$config = (array) config('nativephp.prebuild');
return array_filter($config);

// Then, since config is always an array, you can just loop through it without any extra checks.
```

By itself this is pretty straightforward. Splitting it up in methods and adding more conditional clauses makes it harder to understand what's going on in this case. 

Ofc I might be misunderstanding what it does & again, this based on my personal programming preference.

tests/Unit/Traits/HasPreAndPostProcessingTest.php
```diff
@@ -0,0 +1,49 @@
...
+        ->makePartial();
+
+    $mock->shouldAllowMockingProtectedMethods();
+    $mock->shouldReceive('getPrePostBuildConfig')
+        ->with('prebuild')
+        ->andReturn(collect(Config::get('nativephp.prebuild')));
+
+    $mock->shouldReceive('getPrePostBuildConfig')
+        ->with('postbuild')
+        ->andReturn(collect(Config::get('nativephp.postbuild')));
```
gwleuverink, line 31: Great that this is tested! I see you are actually running the commands to check if files were created. That is great!

There is a preexisting pattern we already use you can do this with which doesn't use mockery (and is a bit more expressive)

This uses a anonymous class that uses the trait you can just call from your test. Since the trait is isolated as is I don't think we need to actually mock any methods.

Bonus: we can still introduce mockery later if we need it

gwleuverink:
Great work on this as always Pete!

I've added two comments, but honestly this comes down to preference I think. If @SRWieZ is okay with it I am too.

I have a question though. I might not be understanding fully how this works so please bear with me.

Isn't this a question of looping through all the prebuild & postbuild commands & running them? The clever `function that returns a function` I saw does seem to overcomplicate it a bit in my eyes. 

Can you elaborate on why it works like this?

src/Traits/HasPreAndPostProcessing.php
PeteBishwhip: A bit of this is cruft. Initially I was going to allow for keys to be used so that this trait can be used in install, build and dev commands.

The usage was going to be like `$this->preProcess('{dev,build,install}')` - That meant that `getPrePostBuildConfig` used to do a lot more than it does now (same with format)

I think another pass at simplifying this is a good idea and I shall do so.

gwleuverink: Yep now it makes sense. Thanks for elaborating 👍🏻

PeteBishwhip:
Updated, refactored and simplified. The scope creep was real.

tests/Unit/Traits/HasPreAndPostProcessingTest.php
```diff
@@ -0,0 +1,44 @@
...
+    unlink($tmpDir.'/postbuild1');
+    unlink($tmpDir.'/postbuild2');
+})
+    ->with([
+        // Empty class with the HasPreAndPostProcessing trait
+        new class
+        {
+            use HasPreAndPostProcessing;
+        },
+    ]);
```
gwleuverink, line 44: beautiful!

gwleuverink (approved):
Perfect! 

One thing to consider. Should we cancel the build when a process fails?

PeteBishwhip:
> Perfect!
> 
> One thing to consider. Should we cancel the build when a process fails?

I did consider this and thought it could be creeping. It may also be that you're fine with things failing in the pre or post steps.

Definitely discussion worthy and can be revisited if decided we should.

SRWieZ (approved):
Tested and approved!

PR #168, opened 2025-02-20
Expand notification api with reference and more events

andreascreten:
### Changes

- Added a reference to the notifications so that you can identify which notification was clicked.

### Added

- Implemented more events for notifications: reply, action, and close.

---

Goes hand in hand with https://github.com/NativePHP/laravel/pull/498

### DIFF
--- a/resources/js/electron-plugin/src/server/api/notification.ts
+++ b/resources/js/electron-plugin/src/server/api/notification.ts
@@ -18,10 +18,13 @@ router.post('/', (req, res) => {
         actions,
         closeButtonText,
         toastXml,
-        event: customEvent
+        event: customEvent,
+        reference,
     } = req.body;
 
     const eventName = customEvent ?? '\\Native\\Laravel\\Events\\Notifications\\NotificationClicked';
+
+    const notificationReference = reference ?? Date.now();
 
     const notification = new Notification({
         title,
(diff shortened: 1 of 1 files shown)


resources/js/electron-plugin/src/server/api/notification.ts
```diff
@@ -18,11 +18,14 @@ router.post('/', (req, res) => {
...
         closeButtonText,
         toastXml,
-        event: customEvent
+        event: customEvent,
+        reference,
     } = req.body;
 
     const eventName = customEvent ?? '\\Native\\Laravel\\Events\\Notifications\\NotificationClicked';
 
+    const notificationReference = reference ?? Date.now();
```
SRWieZ, line 27: Can we add something in case there is two notifications at the same time second ?

Like: `Math.random().toString(36).slice(2, 9)` after `Date.now()`

SRWieZ (changes requested):
I tested the changes, and everything is working as expected. I just encountered an error when sending two notifications one after the other, and they had the same reference.

Next time, try to put your changes in another branch so I can edit myself :)

andreascreten:
> Next time, try to put your changes in another branch so I can edit myself :)

Noted! I pushed the fix for the reference.

SRWieZ (approved):

PeteBishwhip (approved):

PR #169, opened 2025-02-21
fix: update nativephp/laravel version constraint

SRWieZ:
update nativephp/laravel version constraint to allow all versions during development

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -34,7 +34,7 @@
         "php": "^8.1",
         "illuminate/contracts": "^10.0|^11.0",
         "laravel/prompts": "^0.1.1|^0.2|^0.3",
-        "nativephp/laravel": "^1.0@beta",
+        "nativephp/laravel": "*",
         "nativephp/php-bin": "^0.5.1",
         "spatie/laravel-package-tools": "^1.16.4"
     },


gwleuverink (approved):
Some people had trouble updating when this was set to *

This might require further investigation

gwleuverink:
Thanks! I ran into this too 🙏

SRWieZ:
> Some people had trouble updating when this was set to *
> 
> This might require further investigation

I think it's because it was tagged with "*" or I am mistaken ?

PeteBishwhip:
I think its because composer doesn't see the 1.0.0 version as stable. It will still prefer the latest 0.x version over 1.0.0 until a stable tag is released.

SRWieZ:
So, what do you suggest? Should I merge it for now or not?

We can still revert if needed.

PeteBishwhip:
If it's working for you for now, I'm okay with a merge. As you said, its easy to revert.

You will however likely find that when you pull in `nativephp/electron:^1.0@beta`, you'll get a `0.x` version of the laravel package.

SRWieZ:
Then it's okay to merge. In the dev phase, I mostly require `nativephp/electron:dev-main` or another branch.

PR #171, opened 2025-02-21
fix: deepLinkProtocol

SRWieZ:
Fixed a bug that would block the plugin from building.

I followed their documentation here with what was already in place for window management.  
https://www.electronjs.org/docs/latest/tutorial/launch-app-from-url-in-another-app#overview

I couldn't test the feature for now, but at least the electron plugin should build from now on.

### DIFF
--- a/resources/js/electron-plugin/src/index.ts
+++ b/resources/js/electron-plugin/src/index.ts
@@ -90,17 +90,6 @@ class NativePHP {
 
       event.preventDefault();
     });
-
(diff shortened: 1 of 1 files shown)


gwleuverink (approved):

gwleuverink:
Good one! I had this on my list. Thanks for ticking one off 🙏

PeteBishwhip (approved):

PR #176, opened 2025-02-28
Skip errors when copying files during build

gwleuverink:
We ran into an issue while building, caused by a linked storage directory.

This works as expected on mac, but failed on windows. Symlinks do not really exist there, thus the existing conditional clauses had some unexpected behaviour.

Using symlinks is discouraged as per the docs: https://nativephp.com/docs/desktop/1/digging-deeper/files#symlinks

Pete and I had a quick call, and decided to skip (but log) whenever a copy error occurred. 

We might need to harden this further

### DIFF
--- a/src/Traits/CopiesToBuildDirectory.php
+++ b/src/Traits/CopiesToBuildDirectory.php
@@ -10,9 +10,14 @@
 namespace Native\Electron\Traits;
 
 use RecursiveCallbackFilterIterator;
+use Symfony\Component\Filesystem\Path;
+
 use RecursiveDirectoryIterator;
 use RecursiveIteratorIterator;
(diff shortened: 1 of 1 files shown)


PeteBishwhip:
I'd like to get a research task added to our Backlog to look into exactly how differently PHP sees `.lnk` files and how we could effectively capture those as well, but I do agree that for now, we should just warn something could not be copied.

A second thought I had was whether we should document somewhere about the storage directory just to reduce support queries on warnings around it?

Lets see how this does in the wild.

PeteBishwhip (approved):

simonhamp:
This is a good enough solution for now guys. I like it

PR #177, opened 2025-02-28
Copy latest certificate on before build

gwleuverink:
It turns out the `cacert.pem` file was only copied over during the serve command. 
A build before at least running `serve` once was impossible due to the certificate not being present.

This PR adds a step to the build command to move the latest file over on every build.

### DIFF
--- a/src/Commands/BuildCommand.php
+++ b/src/Commands/BuildCommand.php
@@ -14,6 +14,7 @@ use Native\Electron\Traits\LocatesPhpBinary;
 use Native\Electron\Traits\OsAndArch;
 use Native\Electron\Traits\PrunesVendorDirectory;
 use Native\Electron\Traits\SetsAppName;
+use Symfony\Component\Filesystem\Path;
 use Symfony\Component\Process\Process as SymfonyProcess;
 
 use function Laravel\Prompts\intro;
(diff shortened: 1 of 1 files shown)


SRWieZ (commented):
Isn't this handled by the JavaScript side ? Why does npm run dev do it and not npm run build ?

PeteBishwhip:
> Isn't this handled by the JavaScript side ? Why does npm run dev do it and not npm run build ?

It doens't. It's our serve/dev command that copies it. (See php.js)

If you go to build without ever running native:dev/serve, the cacert will not be present and the build will fail.

SRWieZ (approved):
Sorry, my question was: shouldn't we be consistent with the dev command?

I don't really like the current situation
```
"dev": "cross-env node php.js && electron-vite dev --watch",
"build": "electron-vite build",
```

The first step could be to move this PR to a Trait and use it in `DevelopCommand` and `BuildCommand`, and remove it from php.js.

Approved anyway 😉

PeteBishwhip (approved):

gwleuverink:
@SRWieZ you're right. This should be consistent across the board. 

I think it's covered in Pete's follow-up PR

simonhamp:
On the consistency issue: When we build, `php.js` is [executed via `electron-builder.js`](https://github.com/NativePHP/electron/blob/main/resources/js/electron-builder.js#L81)

I suspect the underlying reason for it failing was that the [`certificatePath` variable](https://github.com/NativePHP/electron/pull/178/files#diff-d374a31f5df4418652de60c931f43925923f63b005a28937570c05ae0be3d223L102) wasn't being set on `build`, only on `dev`.

PR #183, opened 2025-03-18
Preserve permissions

XbNz:
`BuildCommand::class`'s `copy` resets all file perms to `100644` r+r for owner, read for everyone else.

Fixed this by just setting it back to original with `chmod`. Tests passing, added new assertions to ensure this works.

There are some intricacies around the 2nd param of `chmod`. It seems to be very user-unfriendly in terms of its behavior when dealing with various permission formats. Decimal values provided as `int` seem to be the way to go, which is also what the `fileperms` function returns by default.

Did a test build and my file permissions came out identical between the real project and the files copied to `resources/app`, I think we're good.

### DIFF
--- a/src/Traits/CopiesToBuildDirectory.php
+++ b/src/Traits/CopiesToBuildDirectory.php
@@ -106,6 +106,14 @@ trait CopiesToBuildDirectory
             } catch (Throwable $e) {
                 warning('[WARNING] '.$e->getMessage());
             }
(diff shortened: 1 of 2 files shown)


simonhamp (approved):

PeteBishwhip (approved):
Brilliant. Thanks for your contribution and thank you for adding a test!

PR #196, opened 2025-03-31
fix: zephpyr.com bundle on Windows

SRWieZ:
The ZIP extension in PHP expects forward slashes (/) as directory separators inside archives, regardless of the operating system.

### DIFF
--- a/src/Commands/BundleCommand.php
+++ b/src/Commands/BundleCommand.php
@@ -283,7 +283,10 @@ class BundleCommand extends Command
                 continue;
             }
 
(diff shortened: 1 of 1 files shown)


simonhamp (approved):

PeteBishwhip (approved):
Looks good to me. I learnt something new today as I didn't know the ZIP extension would require `/` over `\`. I think if you have a known error path for when this affects a ZIP, a test would be good to ensure this doesn't revert back to `\` in the future. Happy for merge.

PR #205, opened 2025-04-05
fix: Please provide a valid cache path.

SRWieZ:
With the added environment variable in this [commit](https://github.com/NativePHP/electron/commit/a2f98889a18d00a5b9e9342f06d51b6c6bf09db6#diff-edf9ec79371541b54bd5647c93b0ba374e50d9986b59c2acd57f410d70d8bb8bR140), we definitely need to create those directories ourselves.

**Why did it work before?**  
Because previously, the app was sometimes writing to itself.  
With bundling, we need every write operation to be outside the app.  
This is why we introduced the usage of `LARAVEL_STORAGE_PATH`.

### DIFF
--- a/resources/js/electron-plugin/dist/server/php.js
+++ b/resources/js/electron-plugin/dist/server/php.js
@@ -24,6 +24,10 @@ const bootstrapCache = join(app.getPath('userData'), 'bootstrap', 'cache');
 const argumentEnv = getArgumentEnv();
 const appPath = getAppPath();
 mkdirpSync(bootstrapCache);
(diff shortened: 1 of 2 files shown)


gwleuverink:
The previous method for providing these directories was by force adding them in the build. But then they would only be created on first run. What you do now is a better solution.

Should we remove these lines from the build command? I don’t think we need them both (untested)

SRWieZ:
We should keep them both for now. I will check later if it's still necessary

simonhamp (approved):

PeteBishwhip (approved):

PR #206, opened 2025-04-05
fix: set LOG_LEVEL to "warning" in production

SRWieZ:
The goal is to not have "info" logs of nativePHP/Electron javascript events.


gwleuverink (approved):
Good call

simonhamp (approved):

PR #207, opened 2025-04-06
fix: handle power monitor events to manage scheduler lifecycle

SRWieZ:
Fixes: https://github.com/NativePHP/laravel/issues/542

Getting a Bad File Descriptor Error after the app is resumed from sleep for every scheduler run.

Laravel scheduler improvement:
* Imported `powerMonitor` module from `electron`
* Added event listeners for `suspend` and `resume` events to manage scheduler state
* Introduced a `stopScheduler` method to clear and reset the scheduler interval


SRWieZ:
This seems like it has fixed my problem.

I don't think it's a breaking change. When the computer goes to sleep, the scheduler wasn't run anyway.

simonhamp (approved):

PeteBishwhip (approved):

PR #208, opened 2025-04-07
fix: App::version() not returning the correct version

SRWieZ:
From electron code comment:
> getVersion(): string; The version of the loaded application. If no version is found in the application's package.json file, the version of the current bundle or executable is returned.


SRWieZ:
Hmm, I think it still works in production. 

But not with `native:serve`. Sooo.. a smaller fix than I thought, but still a fix.

PeteBishwhip (approved):

simonhamp (approved):

PR #210, opened 2025-04-21
fix: take into account all changed keys

SantosVilanculos:
Fix: Account for all changed keys in settingsStore.onDidAnyChange

This commit addresses an issue where only the *first* changed key was being processed when multiple settings were updated simultaneously. The previous implementation used `Object.keys(newValue).find()` which stops after finding the first difference.

The updated code now iterates through *all* keys in the `newValue` object and compares them to their corresponding values in `oldValue`. This ensures that every changed setting triggers the `notifyLaravel` function, providing a complete and accurate representation of settings modifications.

Specifically, this change is crucial in scenarios where:

*   `clear()` or `reset()` are called on the settings store.
*   Multiple settings are updated within a single operation.
*   The application is running in watch mode, where consistent updates are expected.

The `Object.keys(newValue).filter()` method now obtains all changed keys, which are each sent to the `notifyLaravel` to update the backend for each relevant setting.

### DIFF
--- a/resources/js/electron-plugin/dist/server/state.js
+++ b/resources/js/electron-plugin/dist/server/state.js
@@ -2,16 +2,16 @@ import Store from "electron-store";
 import { notifyLaravel } from "./utils.js";
 const settingsStore = new Store();
 settingsStore.onDidAnyChange((newValue, oldValue) => {
-    const changedKey = Object.keys(newValue).find((key) => newValue[key] !== oldValue[key]);
-    if (changedKey) {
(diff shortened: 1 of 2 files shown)


SRWieZ (approved):

SantosVilanculos:
> I'm curious, does the issue still there? Did you found an alternative?

No, I haven't found an alternative. The problem remains.

SantosVilanculos:
I made a new pull request, but I didn't know that deleting a repository fork also closes all open pull requests from it.

New pull request <https://github.com/NativePHP/electron/pull/231>

PR #212, opened 2025-04-26
Add endpoint to relaunch the app

WINBIGFOX:
Added a new `/relaunch` POST endpoint to restart the app with relaunch options. It uses `app.relaunch` with the current arguments and then quits the app. This ensures a seamless restart process for the application.

PR:
NativePHP/laravel: https://github.com/NativePHP/laravel/pull/569

### DIFF
--- a/resources/js/electron-plugin/dist/server/api/app.js
+++ b/resources/js/electron-plugin/dist/server/api/app.js
@@ -6,9 +6,7 @@ router.post('/quit', (req, res) => {
     res.sendStatus(200);
 });
 router.post('/relaunch', (req, res) => {
-    app.relaunch({
-        args: process.argv.slice(1).concat(['--relaunch']),
-    });
+    app.relaunch();
     app.quit();
 });
 router.post('/show', (req, res) => {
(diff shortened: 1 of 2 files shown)


SRWieZ (changes requested):
Why adding `--relaunch` argument ?

It seems unnecessary, and it's there as an example in the docs.

Am I missing something ?

reference: https://www.electronjs.org/docs/latest/api/app#apprelaunchoptions

WINBIGFOX:
Hi @SRWieZ,

Yes, that's exactly what they did in the documentary. I understood that this makes sense because if an application can only have one instance, it can still be restarted in this way. With this argument, it would then be a different instance.

WINBIGFOX:
I just watched the docu again, I may have misunderstood it.
Would probably work reliably even without an argument.

Should I remove the argument again?

SRWieZ:
Yeah, I don't think it's necessary. Let's do without it for now.

WINBIGFOX:
@SRWieZ I have removed the argument 🙂

SRWieZ (approved):

PeteBishwhip (approved):

PeteBishwhip:
@SRWieZ - Both PRs looks good and are approved. I'll leave the merging to you if you're comfortable that the plugin-build workflow is a flap and not an actual issue. Appears to be related to committing the build, so I would assume it's simply going to resolve on the main branch but I leave that to your professional eyes 👀

@WINBIGFOX - Thank you for your contributions! As usual, you're doing amazing stuff

PR #213, opened 2025-04-26
Add auto-updater API to Electron server

WINBIGFOX:
Integrates new auto-updater routes for checking updates and installing them. Also includes event listeners to notify Laravel about update states like availability, download completion, and errors. This enhances the app's ability to manage updates seamlessly.

PR:
NativePHP/laravel: https://github.com/NativePHP/laravel/pull/570


WINBIGFOX:
Oh, okay, I'll check that again.

WINBIGFOX:
@SRWieZ I fixed the test and added the DownloadProgress event

SRWieZ (approved):

resources/js/electron-plugin/dist/server/api/autoUpdater.js
```diff
@@ -0,0 +1,61 @@
+import express from "express";
+import electronUpdater from "electron-updater";
+import { notifyLaravel } from "../utils.js";
+const router = express.Router();
+const { autoUpdater } = electronUpdater;
+router.post("/check-for-updates", (req, res) => {
+    autoUpdater.checkForUpdates();
```
gwleuverink, line 7: Idea: If you add a guard clause here you might be able to prevent the download from happening multiple times. 

Set a Boolean & make sure to reset it when the download finishes/fails

WINBIGFOX: I don't think it's such a bad idea, but I don't know how reliably it will work.

gwleuverink: Yeah it def needs some manual testing. We can always add it later if we want. What do you think @SRWieZ ?

gwleuverink (approved):

SRWieZ (approved):

PR #215, opened 2025-05-01
Adjust window dimensions based on the `resizable` property.

WINBIGFOX:
Updated the logic to set window width and height to ignore stored window state when `resizable` is false. This ensures consistent behavior for fixed-size windows regardless of previous state data.

### DIFF
--- a/resources/js/electron-plugin/dist/server/api/window.js
+++ b/resources/js/electron-plugin/dist/server/api/window.js
@@ -163,7 +163,11 @@ router.post('/open', (req, res) => {
             defaultWidth: parseInt(width),
         });
     }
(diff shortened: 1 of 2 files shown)


WINBIGFOX:
Hey @gwleuverink, are you still missing something in your eyes?
Do you have plans for the next release? There are a few changes I'm already waiting for 🙂

gwleuverink:
I just sent you a dm before reading this

gwleuverink (approved):
Thanks! I ran it and it works 🙏 The terniary was hard to parse, but that was already the case before the change.

I wonder if we can clean this up for readability while we're touching this. Maybe pull it into a function & break the conditionals up?

It was already like this, so it's approved 👍 just recognising a possible refactor opportunity.

simonhamp (approved):

PR #216, opened 2025-05-01
Add support for additional app icon formats

WINBIGFOX:
Extended the InstallsAppIcon trait to copy .ico and .icns icon formats alongside existing ones. This ensures compatibility with platforms requiring these formats.

For macOS: icon.icns
For Windows: icon.ico
Fallback: icon.png

Doc PR: https://github.com/NativePHP/nativephp.com/pull/119

### DIFF
--- a/src/Traits/InstallsAppIcon.php
+++ b/src/Traits/InstallsAppIcon.php
@@ -8,6 +8,10 @@ trait InstallsAppIcon
     {
         @copy(public_path('icon.png'), __DIR__.'/../../resources/js/build/icon.png');
         @copy(public_path('icon.png'), __DIR__.'/../../resources/js/resources/icon.png');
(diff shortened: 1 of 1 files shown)


gwleuverink (changes requested):
Looks good 👍 

Can you verify we don’t need to add these here too: https://github.com/NativePHP/electron/blob/main/src/Traits/InstallsAppIcon.php

If it isn’t coupled it might be a smell we can remove one of the two.

If it is, it wouldn’t be a bad idea to leave a comment that these two places must be kept in sync manually (or perhaps decouple it by using the internal config?)

WINBIGFOX:
@gwleuverink I don't know if I'm misunderstanding something here, but I made the change to the file you referred to

gwleuverink:
No you're right I was going over PR's and must have gotten some wires crossed. Approved!

gwleuverink (approved):

SRWieZ (approved):

PR #222, opened 2025-05-18
feat: add electron-builder command to publish electron-builder.js

blumewas:
Add `native:electron-builder` command to make the `electron-builder.js` file publishable to our resources directory. This would solve my need as a developer to make changes to backend building step.

---

The command creates a copy of the `electron-builder.js` provided by installing this package inside the `resources/vendor/nativephp/js`. After that I will create a backup of the `package.json` and updates the scripts section so that our copy will be used when building.

When running the command with the `--restore` option the `package.json` will be restored from a backup.

### DIFF
--- a/src/Commands/ElectronBuilderCommand.php
+++ b/src/Commands/ElectronBuilderCommand.php
@@ -1,0 +1,26 @@
+<?php
+
+namespace Native\Electron\Commands;
+
+use Illuminate\Console\Command;
+use Native\Electron\Traits\PublishesElectronBuilderJs;
+
+class ElectronBuilderCommand extends Command
+{
+    use PublishesElectronBuilderJs;
+
+    protected $signature = 'nativephp:electron-builder {--force} {--restore}';
+
+    public function handle(): void
+    {
+        if ($this->option('restore')) {
+            // Restore the electron-builder.js file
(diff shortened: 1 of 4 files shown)


simonhamp:
@blumewas can you describe more specifically what you're modifying in the `electron-builder.js` and why? Perhaps we can more directly support your use-case without introducing what feels like a bit of a hack

blumewas:
Hey @simonhamp,
There was PR asking for addition of one line for the code. You pointed out the fact that is not needed for everybody and only a set of people using Native PHP. I wanted to create a possibility to achieve this. 

I also wrote a command merging parts of the native-php config file into the returned JS of the electron-builder but found that part rather complex. So I came up with this simpler solution.

To conclude I would say that I am trying to give devs a way to customize the electron-builder config for some use cases that are not required for everybody. If this feels kinda hacky to you which it is because it may only be used if you know what to do I understand if you would reject it.

simonhamp:
What specific changes are _you_ trying to make to _your_ `electron-builder.js`?

blumewas:
Currently I had no chance to dig that deep into the project.

The PR was premature so I’ll close it. For me personally there is NO need atm.

PR #227, opened 2025-06-08
fix: symfony console deprecation notice

SRWieZ:
My console was spamming this message, and it's annoying.

Capture d’écran 2025-06-08 à 16 49 52


PeteBishwhip (approved):

gwleuverink:
It looks good to me. It's the same fix applied on the Laravel repo.

I wonder though, since this is a Laravel command, if they'll fix this downstream. The docs still say the api is the same on v12

gwleuverink (approved):

PR #238, opened 2025-07-29
use vendor cacert.pem for custom binaries

gwleuverink:
When using custom PHP binaries the `CopiesCertificateAuthority` trait would look for the `cacert.pem` file relative to the custom binary path. 

It's not documented that users should provider their own certificate bundle in this case. I see no reason why we should not use the default certs used by `php-bin`. Since we include the repo by default it should always be there & be updated.

This PR removes the conditional & makes sure the certs in `vendor/nativephp/php-bin/cacert.pem` are always used.

Fixes: https://github.com/NativePHP/laravel/issues/655

### DIFF
--- a/src/Traits/CopiesCertificateAuthority.php
+++ b/src/Traits/CopiesCertificateAuthority.php
@@ -17,13 +17,6 @@ trait CopiesCertificateAuthority
 
             $phpBinaryDirectory = base_path('vendor/nativephp/php-bin/');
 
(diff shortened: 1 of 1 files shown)


gwleuverink:
Alternatively we could search both locations for a `cacert.pem` file. But thinking about this I'm not sure about the benefits.

SRWieZ (approved):
I agree

gwleuverink:
@NativePHP/contributors one more review to push this through 👀

PR #241, opened 2025-08-01
remove workingDirectory from Windows deep-link handler

gwleuverink:
Deeplink handling on Windows was passing a `workingDirectory` parameter to the `OpenedFromURL` event.

This was causing issues when opening deep-links on Windows

Fixes: https://github.com/NativePHP/laravel/issues/599

### DIFF
--- a/resources/js/electron-plugin/dist/index.js
+++ b/resources/js/electron-plugin/dist/index.js
@@ -149,7 +149,6 @@ class NativePHP {
                             event: "\\Native\\Laravel\\Events\\App\\OpenedFromURL",
                             payload: {
                                 url: commandLine[commandLine.length - 1],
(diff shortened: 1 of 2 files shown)


SRWieZ (commented):
Can't we change `OpenedFromURL` to include this directory as a nullable parameter? I believe it's useful information to have. I plan to use it in one of my project-based projects, such as when I run `phpstorm` in a project directory and it opens the project directly.

SRWieZ (approved):

SRWieZ:
After reading this again, I approve!

https://www.electronjs.org/docs/latest/tutorial/launch-app-from-url-in-another-app

simonhamp (approved):

PR #244, opened 2025-08-06
Fix port conflicts when running multiple NativePHP apps on Windows

gwleuverink:
Multiple NativePHP applications cannot run simultaneously on Windows due to port binding conflicts. The second app fails with exit code 4294967294 (`EADDRINUSE`) and `net::ERR_NAME_NOT_RESOLVED` errors.

See issue https://github.com/NativePHP/laravel/issues/651

`getPort` sometimes reports ports as available when they're actually bound by another process. This happens due to Windows-specific port detection issues, including `TIME_WAIT` states and reserved port ranges that get-port doesn't properly detect.

## Proposed fix

Enhanced getPhpPort() to validate that suggested ports are actually bindable:

1. Fast path: Try get-port first (works in 99% of cases)
2. Validation: Test binding to the suggested port with a real TCP server
3. Smart fallback: If validation fails, incrementally search from suggestedPort + 1 upward

### DIFF
--- a/resources/js/electron-plugin/src/server/php.ts
+++ b/resources/js/electron-plugin/src/server/php.ts
@@ -8,6 +8,7 @@ import {promisify} from 'util'
 import {join} from 'path'
 import {app} from 'electron'
 import {execFile, spawn, spawnSync} from 'child_process'
+import {createServer} from 'net'
 import state from "./state.js";
 import getPort, {portNumbers} from 'get-port';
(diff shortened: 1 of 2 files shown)


gwleuverink:
@NativePHP/contributors What do you guys think about this? 

It's not the most elegant approach, but I think it's pretty solid & will handle the 1% scenario where `getPort` isn't working as expected

PR #253, opened 2025-09-07
Allow overwriting of default web preferences

JulianGlueck:
**This PR will allow the user to overwrite NativePHP's default web preferences passed to Electron.**

At the time of creating this PR, these are `backgroundThrottling`, `spellcheck`, `preload`, `sandbox`, `contextIsolation`, and `nodeIntegration`. See details here: https://github.com/NativePHP/laravel/issues/688

### DIFF
--- a/resources/js/electron-plugin/src/server/api/window.ts
+++ b/resources/js/electron-plugin/src/server/api/window.ts
@@ -302,8 +302,8 @@ router.post('/open', (req, res) => {
         autoHideMenuBar,
         ...(process.platform === 'linux' ? {icon: state.icon} : {}),
         webPreferences: {
-            ...webPreferences,
-            ...defaultWebPreferences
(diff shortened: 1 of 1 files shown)


gwleuverink:
Heya! I took a mo to make the preload script work with enhanced security preferences https://github.com/NativePHP/electron/pull/255

I think we should reassess which of the default preferences may be overwritten without breaking the setup.

Right now in my PR it needs to force the `preload`, `contextIsolation` & `sandbox` preferences. I think the other stuff may be overwritten without issues

gwleuverink:
I've shuffled some things around and added this in #255. Closing this in favor of that PR

Added you as a co-author here: https://github.com/NativePHP/electron/pull/255/commits/a1061ff92f53aa4ed67cb18c5398622c51696ff4

PR #255, opened 2025-09-09
Better webPreferences defaults

gwleuverink:
This PR improves the default security preferences & changes `preload.js` to support those changes:

- disabled node integration
- enabled context isolation (required - can't be overwritten)
- refactored preload script to use the contextBridge API
- added `native:init` event for registering Native listeners (prevents race conditions where preload hasn't evaluated yet)
- updated which of these defaults may be overwritten (`preload`, `contextIsolation` and `sandbox` cannot be changed)
- make sure Window & MenuBar use the same preferences
- added the ability to pass custom webPreferences to MenuBar windows

See https://github.com/NativePHP/laravel/issues/688 for more details

--- 

This PR also introduces a event that's called whenever the `preload` script is fully evaluated. In the past some people have reported race conditions where they register a listener with `Native.on()` but the Native object is not available yet.

This event adresses that:

``` js
document.addEventListener('native:init' function() {

    Native.on("Native\\Laravel\\Events\\Windows\\WindowBlurred", (payload, event) => {
        //
    });
})
```

### DIFF
--- a/resources/js/electron-plugin/dist/preload/index.mjs
+++ b/resources/js/electron-plugin/dist/preload/index.mjs
@@ -51,7 +51,7 @@ ipcRenderer.on('native-event', (event, data) => {
         });
     }
 });
-contextBridge.exposeInMainWorld('_native_init', (function () {
+contextBridge.exposeInMainWorld('native:initialized', (function () {
     window.dispatchEvent(new CustomEvent('native:init'));
-    delete window['_native_init'];
+    return true;
 })());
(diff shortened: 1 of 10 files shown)


gwleuverink:
Additionally I've deleted the preload file at `src/preload/index.js` (also from vite builder config)

We're only using the one from the electron plugin, so the one I removed was probably an artifact. Everything is still working as expected in my testing environment. 

Can someone confirm this is okay? I've checked high and low but that second preload wasn't doing anything at all.

gwleuverink:
I've locked `sandbox`, `preload` & `contextIsolation` prefs so they can't change. Other defaults may be overwritten now.

While working on this I noticed we couldn't pass custom webPreferences to MenuBar windows. I've added this in https://github.com/NativePHP/laravel/pull/694

SRWieZ (approved):

simonhamp (approved):

SRWieZ:
Already merged in https://github.com/NativePHP/desktop

PR #259, opened 2025-10-22
Adding extra check for Node.js running under Rosetta 2 for Apple Silicon

### LINKED ISSUE #22: Call to undefined function mb_split() - Missing mbstring extension in bundled PHP binary on macOS
### What were you trying to do?

I was trying to run a fresh NativePHP application on macOS

### What happened?

```
Error in PHP:
  Call to undefined function Illuminate\Support\mb_split() (0) in ~/Projects/mixlab/vendor/laravel/framework/src/Illuminate/Support/Str.php:1685
```
The issue is that NativePHP's bundled PHP binary (from nativephp/php-bin) is missing the mbstring extension, which is required by Laravel's Str class.

### How to reproduce the bug

1. Install Laravel `laravel new app`
2. composer require nativephp/desktop
3. php artisan native:install 
4. php artisan native:run

### Debug Output

```
{
    "Environment": {
        "PHP": {
            "Version": "8.3.26",
            "Path": "/Users/irongs/Library/Application Support/Herd/bin/php83"
        },
        "Laravel": {
            "Version": "11.46.1",
            "ConfigCached": false,
            "RoutesCached": false,
            "DebugEnabled": true
        },
        "Node": {
            "Version": "v22.17.1",
            "Path": "/Users/irongs/Library/Application Support/Herd/config/nvm/versions/node/v22.17.1/bin/node"
        },
        "NPM": {
            "Version": "11.4.2",
            "Path": "/Users/irongs/Library/Application Support/Herd/config/nvm/versions/node/v22.17.1/bin/npm"
        },
        "OperatingSystem": "Darwin"
    },
    "NativePHP": {
        "Versions": {
            "nativephp/desktop": "2.0.1.0",
            "nativephp/php-bin": "1.1.1.0"
        },
        "Configuration": {
            "Provider": "App\\Providers\\NativeAppServiceProvider",
            "BuildHooks": {
                "Pre": [],
                "Post": []
            },
            "NotarizationEnabled": false,
            "AzureTrustedSigningEnabled": false,
            "CustomPHPBinary": false
        }
    }
}
```

### Which operating systems have you seen this occur on?

macOS

### Notes

The bundled PHP binary is extracted from /vendor/nativephp/php-bin/bin/mac/x86/php-8.3.zip and is missing the mbstring extension. When running php artisan tinker with Herd's PHP, mb_split() works fine, confirming the issue is specific to NativePHP's bundled binary.
I verified this by directly testing the bundled PHP binary:
cd /Users/irongs/Projects/mixlab/vendor/nativephp/desktop/resources/build/php
./php ../../../../../../artisan tinker

```
Output:
Psy Shell v0.12.12 (PHP 8.3.15 — cli) by Justin Hileman
> mb_split()
   Error  Call to undefined function mb_split().
```
This confirms that the bundled PHP binary (PHP 8.3.15) does not have the mbstring extension compiled/enabled, while the same function works correctly with Herd's PHP installation.

cemarta7:
I was able to fix [issue 22](https://github.com/NativePHP/desktop/issues/22) if runnign under Rosetta2 we need to do one more check.


gwleuverink:
We need to verify this is the same issue as https://github.com/NativePHP/desktop/issues/22

I suspect your problem is caused by running npm under rosetta on a Apple Sillicon machine.

The OP of the linked issue is on a Intel machine and correctly using the x86 binary. 

You are both seeing the same error, but perhaps for different reasons:

- there may be an issue with the mac x86 binary (you both see this error)
- you should not be using the x86 binary int the first place

gwleuverink:
Could you check the output of these commands for me?

```
# Check which architecture your terminal is using
arch

# Check Node's architecture
node -p "process.arch"

# Check nvm's default architecture
nvm current
```

gwleuverink:
Can you check if the fix here does anything for you?

https://github.com/NativePHP/desktop/issues/22#issuecomment-3442644326

I expect it will then build with the correct `x86` binary. Your error would disappear, but you'd still run it under rosetta under the hood.

I think this might only be the case when running `native:run` where it's arch is mismatched due to running node under rosetta. But when building the architecture is passed in, not detected. So in theory your build should use the correct binary.

Can you verify that hypothesis?

